PR #1, opened 2021-12-09
Co-registered S1 SLC schema

vbrancat:
This (small) PR introduces a schema for the S1-A/B co-registered SLC workflow. Some of the decisions made in designing the schema will drive/affect the design of the workflow.

Some food for thoughts:
1. As an input file, the workflow should take the path to reference and secondary granules. If a list of `burst_id` is provided, the workflow will process only that burst IDs. If `burst_id` is empty, all the bursts in the zip files are processed.
2. Orbits are not required. If left empty, the workflow will download the required orbits (working on a script)
3. DEM file is not required. If not assigned, DEM should be downloaded automatically. Thinking to write a wrapper to download the Copernicus DEM.

Please, double-check if the selected directory structure makes sense.

### DIFF
--- a/defaults/cslc_s1.yaml
+++ b/defaults/cslc_s1.yaml
@@ -1,0 +1,34 @@
+runconfig:
+   name: cslc_s1_workflow_default
+
+   groups:
+       pge_name_group:
+           pge_name: CSLC_S1_PGE
+
+       input_file_group:
+           # Required if burst_id is not provided. List of reference and secondary S1 ZIP files
+           input_file_path:
+           # List of reference and secondary orbit file
+           orbit_file_path:
+           # List of unique burst IDs. Required if input_file_path is not provided
+           burst_id:
+
+       dynamic_ancillary_file_group:
+           # Digital Elevation Model
+           dem_file:
+
+       product_path_group:
+           # Directory where PGE will place results
+           product_path:
+           # Directory where SAS can write temporary data
+           scratch_path:
+           # Intermediate file name. SAS writes the output to this file.
+           # PGE ay rename the product according to file naming convention
+           sas_output_file:
+
+       primary_executable:
+           product_type: CSLC_S1
+
+       processing:
+           rdr2geo:
+               lines_per_block: 1000
--- a/defaults/cslc_s1.yaml
+++ b/defaults/cslc_s1.yaml
@@ -14,7 +14,7 @@ runconfig:
            burst_id:
            # Options for reference burst
            reference_burst:
-               # Required. Flag indicating reference burst
+               # Required. Flag indicating reference burst. If True, the input zip file is a reference date
                is_reference: True
                # File path to reference burst to which coregister secondary burst.
                # Only used if the flag is_reference is set to False.
--- a/defaults/cslc_s1.yaml
+++ b/defaults/cslc_s1.yaml
@@ -14,7 +14,10 @@ runconfig:
            burst_id:
            # Options for reference burst
            reference_burst:
+               # Required. Flag indicating reference burst
                is_reference: True
+               # File path to reference burst to which coregister secondary burst.
+               # Only used if the flag is_reference is set to False.
                file_path:
 
        dynamic_ancillary_file_group:
--- a/defaults/cslc_s1.yaml
+++ b/defaults/cslc_s1.yaml
@@ -6,12 +6,16 @@ runconfig:
            pge_name: CSLC_S1_PGE
 
        input_file_group:
-           # Required. List of reference and secondary S1 ZIP files
-           input_file_path:
-           # List of reference and secondary orbit file
+           # Required. List of SAFE files (min=1)
+           safe_file_path:
+           # Required. List of orbit (EOF) files (min=1)
            orbit_file_path:
-           # List of unique burst IDs. Required if input_file_path is not provided
+           # List of (unique) burst IDs to process
            burst_id:
+           # Options for reference burst
+           reference_burst:
+               is_reference: False
+               file_path:
 
        dynamic_ancillary_file_group:
            # Digital Elevation Model
@@ -32,3 +36,10 @@ runconfig:
        processing:
            rdr2geo:
                lines_per_block: 1000
+
+           geo2rdr:
+               lines_per_block: 1000
+
+           resample:
+               lines_per_block: 1000
+
--- a/defaults/cslc_s1.yaml
+++ b/defaults/cslc_s1.yaml
@@ -6,7 +6,7 @@ runconfig:
            pge_name: CSLC_S1_PGE
 
        input_file_group:
-           # Required if burst_id is not provided. List of reference and secondary S1 ZIP files
+           # Required. List of reference and secondary S1 ZIP files
            input_file_path:
            # List of reference and secondary orbit file
            orbit_file_path:
--- a/defaults/cslc_s1.yaml
+++ b/defaults/cslc_s1.yaml
@@ -14,7 +14,7 @@ runconfig:
            burst_id:
            # Options for reference burst
            reference_burst:
-               is_reference: False
+               is_reference: True
                file_path:
 
        dynamic_ancillary_file_group:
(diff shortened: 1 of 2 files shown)


defaults/cslc_s1.yaml
```diff
@@ -0,0 +1,34 @@
+runconfig:
+   name: cslc_s1_workflow_default
+
+   groups:
+       pge_name_group:
+           pge_name: CSLC_S1_PGE
+
+       input_file_group:
+           # Required. List of reference and secondary S1 ZIP files
```
hfattahi, line 9: do we really need both reference and secondary zip files to coregister every burst? I think it depends on the design. Since we are at the very beginning I suggest that we avoid requiring the reference zip file every time we coregister a secondary burst to a reference burst.

defaults/cslc_s1.yaml
```diff
@@ -0,0 +1,34 @@
...
+           input_file_path:
+           # List of reference and secondary orbit file
+           orbit_file_path:
+           # List of unique burst IDs. Required if input_file_path is not provided
```
hfattahi, line 13: why burst IDs required only when we don't have the input files? Can we do anything if there is no input file anyways?

schemas/cslc_s1.yaml
```diff
@@ -0,0 +1,51 @@
...
+        primary_executable:
+           product_type: enum('CSLC_S1')
+
+        # This section includes parameters to tweak the workflow
+        processing:
+           # Polarization to process. If not set, process all polarizations for that burst
+           polarization: any(list('HH', 'HV', 'VH', 'VV'), min=1, max=4), str(min=2, max=2), required=False)
+
+           # Options to run rdr2geo
+           rdr2geo: include('rdr2geo_options', required=False)
```
hfattahi, line 38: need geo2rdr and resample options as well

hfattahi (commented):
Thanks @vbrancat for starting with this PR which will help to make sure we have a clear design before implementation. I'm a little bit unclear on the exact design. I assume this is a schema for S1 coregistered SLC in radar coordinates only. Is that correct?
I put in few comments and I will start a discussion which may help me better understand what the overall workflow will be.

defaults/cslc_s1.yaml
```diff
@@ -0,0 +1,45 @@
...
+           # Required. List of SAFE files (min=1)
+           safe_file_path:
+           # Required. List of orbit (EOF) files (min=1)
+           orbit_file_path:
+           # List of (unique) burst IDs to process
+           burst_id:
+           # Options for reference burst
+           reference_burst:
+               is_reference: False
+               file_path:
```
LiangJYu, line 18:
```suggestion
               is_reference: True 
               file_path:
```
What do you think about making the default `is_reference` to `True` and assume any user starting from nothing? Also allows us to not have to worry about what's the correct default `file_path` for non-reference cases.

schemas/cslc_s1.yaml
```diff
@@ -0,0 +1,74 @@
...
+        pge_name_group:
+           pge_name: enum('CSLC_S1_PGE')
+
+        input_file_group:
+           # Required. List of SAFE files (min=1)
+           safe_file_path: list(str(), min=1)
+           # Required. List of orbit (EOF) files
+           orbit_file_path: list(str(), min=1)
+           # List of (unique) burst ID to process
+           burst_id: list(str(), min=0, max=27, required=False)
```
LiangJYu, line 14: `max=27` is for all bursts in a swath? Are there cases where there's more than 9 bursts per sub-swath? Will other Sentinel 1 modes (besides IW) be used for CSLC? If so does the cap make work for those modes?

schemas/cslc_s1.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+    groups:
+        pge_name_group:
+           pge_name: enum('CSLC_S1_PGE')
+
+        input_file_group:
+           # Required. List of SAFE files (min=1)
+           safe_file_path: list(str(), min=1)
+           # Required. List of orbit (EOF) files
+           orbit_file_path: list(str(), min=1)
```
LiangJYu, line 12: What's the use case for more than 1 `orbit_file_path`?

schemas/cslc_s1.yaml
```diff
@@ -0,0 +1,74 @@
...
+           geo2rdr: include('rdr2geo_geo2rdr_shared_options', required=False)
+
+           # Options to run resample
+           resample: include('resample_options', required=False)
+
+
+# Group of processing options
+reference_burst_options:
+    # Required. Flag indicating reference burst
+    is_reference: bool()
```
LiangJYu, line 52:
```suggestion
    # Required. If True, rdr2geo will be performed on the SAFE files and burst IDs above
    is_reference: bool()
```
Documentation updated to account for a case where: for a given ROI made up of a collection of burst IDs, on the day chosen as reference, burst IDs span 2 swaths. If reference days are chosen to avoid the condition I described, then this comment maybe ignored..

schemas/cslc_s1.yaml
```diff
@@ -0,0 +1,74 @@
...
+    # File path to reference burst
```
LiangJYu, line 53:
```suggestion
    # File path to reference burst to process secondary bursts against. Only used if if_refernce above is False.
```
Suggestion for clarity. Apologies if nomenclature is off.

LiangJYu (commented):
To confirm we're all on the same page: The schema reflects a single executable that chooses a subworkflow based on the `is_reference`. If `is_reference` is `True`  then given SAFE zips+bursts IDs will be processed through topo. If `is_reference` is `False`  then given SAFE zips+bursts IDs will be processed through geo2rdr+resample. 
```python
def CSLC_burst_processing():
    if is_reference:
        rdr2geo(zips, burst_ids)
    else:
        geo2rdr(zips, burst_ids, reference_file_path)
        resample_slc(zips, burst_ids)
```

defaults/cslc_s1.yaml
vbrancat: Sounds good to me. I will apply this change

schemas/cslc_s1.yaml
vbrancat: Good point. To my knowledge, typically each sub-swath has 9 bursts. However, I was not able to confirm that number consulting official Sentinel pages. I am inclined to remove the cap on the max number of bursts to process

schemas/cslc_s1.yaml
vbrancat: We have specified the `safe_file_path` as a list of file paths to SAFE files (e.g. more than one date). If this is correct, you want to assume that you will have more orbit files. Am I missing something?

schemas/cslc_s1.yaml
vbrancat: Can you elaborate more what you mean?

schemas/cslc_s1.yaml
LiangJYu: I forgot we're not supplying the orbit file(s). My bad.

schemas/cslc_s1.yaml
LiangJYu: I'm trying to account for the following scenario:

Reference ROI spans more than 1 swath: ROI includes IDs: 1, 2, 3. IDs 1 and 2 are in swath N and  and ID 3 is in swath N+1. For the YAML capture this, `safe_file_path` will have more than item, (swath N and swath N+1) and `is_reference` is `True`.

hfattahi: @LiangJYu I think your suggestion seems a bit too detail for an end user who may want to only deal with the run config. How about ```Flag indicating reference burst. If True the input zip file is a reference date```. What Virginia has  also is clear to me.

hfattahi (approved):
Thanks @vbrancat. This looks good to me and seems consistent with what we discussed [here](https://github.com/opera-adt/COMPASS/discussions/3)

schemas/cslc_s1.yaml
LiangJYu: Sounds good to me.

LiangJYu (approved):

LiangJYu:
LGTM

PR #4, opened 2022-01-05
rdr2geo module

LiangJYu:
This PR introduces:

* rdr2geo processing code
* runconfig lass
* auxiliary classes and functions in yaml_argparse.py and helper.py

This PR addresses:

* syntax errors in schema
* minor cleanup in schema and defaults

TODO:
* ~~test; as-is, beyond basic linting and formatting, nothing has been tested~~
* unit tests
* ~~add installer~~
* more documentation


src/COMPASS/runconfig.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            raise FileNotFoundError
+
+        # validate yaml file taken from command line
+        try:
+            yamale.validate(schema, data)
+        except yamale.YamaleError as yamale_err:
+            err_str = f'Validation fail for {workflow_name} runconfig yaml {yaml_path}.'
+            logging.error(err_str)
+            raise yamale.YamaleError(err_str) from yamale_err
```
hfattahi, line 83: would this provide a meaningful error message on why validation fails (in case if it fails)?

Could you use something like [this]( 
https://github.com/23andMe/Yamale#:~:text=try%3A%0A%20%20%20%20yamale.validate(schema%2C%20data)%0A%20%20%20%20print(%27Validation%20success!%20%F0%9F%91%8D%27)%0Aexcept%20YamaleError,error%20in%20result.errors%3A%0A%20%20%20%20%20%20%20%20%20%20%20%20print(%27%5Ct%25s%27%20%25%20error)%0A%20%20%20%20exit(1)) to print out the errors?

src/COMPASS/runconfig.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # load default config
+        parser = YAML(typ='safe')
+        default_cfg = f'{helpers.WORKFLOW_SCRIPTS_DIR}/defaults/{workflow_name}.yaml'
+        with open(default_cfg, 'r') as f_default:
+            cfg = parser.load(f_default)
```
hfattahi, line 89: Nice! I assume this contains the default parameters of the configuration.

src/COMPASS/runconfig.py
```diff
@@ -0,0 +1,144 @@
...
+
+        with open(yaml_path, 'r') as f_yaml:
+            user = parser.load(f_yaml)
```
hfattahi, line 92: user_cfg maybe?

src/COMPASS/runconfig.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # copy user suppiled config into default config
+        helpers.deep_update(cfg, user)
+
+        validate_group(cfg['groups'])
```
hfattahi, line 97: Can you add a comment, what does this function do?

hfattahi (commented):
Thank you @LiangJYu for this PR. Overall looks good. I put in some minor comments.
One concern I have is that currently you are not allowing to pass multiple safe files. Since the processing units are bursts anyways, I guess we could probably loop over multiple safe files. We need to make a decision on the design for cases where user is asking for many bursts that may be covered by multiple safe files. If we clarify this part, then the rest seems good to me. The run configuration parser design seems consistent and comprehensive enough to accommodate next steps (geo2rdr and resample). I did not run the code on real data but the rdr2geo seems consistent with our several rounds of offline prototypes and should work.  
And we should decide on unit test which is on your todo list anyways.

src/COMPASS/rdr2geo.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+if __name__ == "__main__":
+    '''run rdr2geo from command line
+    '''
+    # load command line args
+    rdr2geo_parser = YamlArgparse()
+    # get a runconfig dict from command line args
+    rdr2geo_runconfig = RunConfig(rdr2geo_parser.args)
+    # run rdr2geo
```
hfattahi, line 77: For clarity I suggest that at this point we write the cfg to disk, documenting exact parameters used for running geo2rdr. 
It would be nice if we write cfg in the same format of our schema. 
The use case would be that a user may only pass a very short run config file to the SAS but always a complete run config (combination of dafault and user defined parameters) is written to disk to clarify what has been  used for processing. @yunjunz has nicely implemented this in mintpy.

src/COMPASS/helpers.py
```diff
@@ -0,0 +1,73 @@
...
+import pathlib
+
+from osgeo import gdal
+import h5py
+
+def check_file_path(file_path: str) -> None:
+    '''check if file path exists and raise error it does not'''
+    if not os.path.isfile(file_path):
+        err_str = f'{file_path} not found'
+        logging.error(err_str)
```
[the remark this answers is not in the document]
vbrancat: I would propose to use `journal `as it would come for free from the `isce3` dependencies. I am not sure if we have any requirement on log and error formatting and how flexible `journal` is on that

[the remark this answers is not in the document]
LiangJYu: I have no preference. I nice part of using journal is that'll encourage more pyre use (for future workflow management?). Just need somebody to learn or who's already adept...

src/COMPASS/rdr2geo.py
```diff
@@ -0,0 +1,78 @@
...
+    ellipsoid = proj.ellipsoid
+
+    # process rdr2geo for found bursts
+    for i, burst in enumerate(bursts):
+        # get isce3 objs from burst
+        rdr_grid = burst.as_isce3_radargrid()
+        isce3_orbit = burst.get_isce3_orbit(cfg.orbit_path)
+
+        # init output directory
+        output_path = f'{cfg.sas_output_file}/{burst.burst_id}'
```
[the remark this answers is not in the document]
LiangJYu: This is carryover behavior from ISCE/NISAR. 😅 I'll rework to take advantage of optional topo layers.

src/COMPASS/runconfig.py
LiangJYu: As-is It will dump out errors in schema validation. 
image

LiangJYu:
@yunjunz As briefly discussed in today's meeting, I've been unsuccessful in my efforts to copy [defaults](https://github.com/LiangJYu/COMPASS/tree/rdr2geo/defaults) and [schemas](https://github.com/LiangJYu/COMPASS/tree/rdr2geo/schemas) to my Conda environment in the install process. 7762842 contains my unsuccessful attempt to install with `package_data`. May I trouble to you take a look and check for anything wrong or missing? Thanks.

The following is the directory structure I'd like to get. Highlights show pieces I manually copied.
image

src/COMPASS/util/runconfig.py
```diff
@@ -0,0 +1,167 @@
...
+    # Check 'product_path_group' section of runconfig.
+    # Check that directories herein have writing permissions
+    product_path_group = group_cfg['product_path_group']
+    helpers.check_write_dir(product_path_group['product_path'])
+    helpers.check_write_dir(product_path_group['scratch_path'])
+    helpers.check_file_path(product_path_group['sas_output_file'])
+
+    # Check polarizations to process.
+    if 'polarization' not in group_cfg['processing']:
+        group_cfg['processing']['polarization'] = ['HH', 'HV', 'VH', 'VV']
```
vbrancat, line 56: Note that Sentinel is not a quad-pol but a dual-pol system. May be, if no polarization is found either we read them from the SAFE file or by default we process the most used pol for OPERA (VV, VH). However, this m
s1-constellation-mode-polarization-observation-geometry-800
ay cause troubles at the North and South pol where the dual-pol mode used is HH and HV. See pic from ASF website

hfattahi: How about we check the file name of the zip file and extract polarizations if not given?
The file naming convention for S1 is [here](https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/naming-conventions). For example :
S1B_IW_SLC__1S**DV**_20211127T134403_20211127T134430_029774_038DC7_08D0.zip

**DV** represents Dual Pol V, i.e., VV and VH

The different possibilities are:

SH (single HH polarisation)
SV (single VV polarisation)
DH (dual HH+HV polarisation)
DV (dual VV+VH polarisation)

LiangJYu:
@rtburns-jpl May I trouble you for assistance with this issue with copying [`package_data`](https://github.com/opera-adt/COMPASS/pull/4#issuecomment-1006232233)? Thanks.

yunjunz:
@LiangJYu I would recommend the following for the code structure:
1. using lower case for the module name, as `src/compass`
2. use the plural form of `utils`, as `src/compass/utils` for the sub-module name. This is similar to `docs` instead of `doc` in other packages.
3. maybe put `defaults` and `schemas` under `compass` directly, instead of under `compass/utils`?

Regarding the packaging of data files using `setup.cfg`, `include_package_data` and `MANIFEST.in` works together; while `package_data` is independent for extra stuff (https://setuptools.pypa.io/en/latest/userguide/datafiles.html). I have updated the `MANIFEST.in` and `setup.cfg` file accordingly, based on the proposed code structure above. Local testing passed on my laptop (data file is in the `site-packages` dir).

I did not push changes related to the file moving, in case you might has some uncommited local changes on your side. However, I could push those changes as well, let me know pls.

LiangJYu:
@yunjunz Thank you for solving this! Please push your changes at your earliest convenience.

yunjunz:
My changes are now pushed. However, the `COMPASS/__init__.py` and `COMPASS/rdr2geo.py` seem to be still in the old structure here, and not recognized from my local fork. @LiangJYu could you change them on your side? You would need to update the modules as well, due to the structure change. 

Related to the module import, I would recommend avoiding relative imports, and using the absolute imports always.

hfattahi:
@LiangJYu, @vbrancat I forgot to ask what is the design for the top-level workflow. Let's assume SDS whats to run coreg-slc. Are we going to have one coregister_sentinel1.py that would internally know to run rdr2geo or "geo2rdr+resample" based on the is_reference flag? or are you considering something else?

vbrancat:
@hfattahi, this is clarified in the discussion on the design of CSLC workflow in radar coordinates. I remember that we all agreed that is not convenient for SDS to have 2 separate workflows (one for `rdr2geo` if reference burst and another for `geo2rdr`+`resample` for secondary bursts). If I recollect correctly, we settled on having one workflow internally dealing with these two cases. The latter is the design we are shooting for. Am I missing something?

hfattahi:
> @hfattahi, this is clarified in the discussion on the design of CSLC workflow in radar coordinates. I remember that we all agreed that is not convenient for SDS to have 2 separate workflows (one for `rdr2geo` if reference burst and another for `geo2rdr`+`resample` for secondary bursts). If I recollect correctly, we settled on having one workflow internally dealing with these two cases. The latter is the design we are shooting for. Am I missing something?

Yes that is correct. I did not see the workflow in this PR. It would be great to add it then.

src/compass/utils/runconfig.py
```diff
@@ -0,0 +1,228 @@
...
+from itertools import cycle
+import os
+from types import SimpleNamespace
+
+import journal
+import yamale
+from ruamel.yaml import YAML
+
+from compass.utils import helpers
+from compass.utils.wrap_namespace import wrap_namespace
```
vbrancat, line 12: It seems that this function is not contained in the `utils` folder. Make sure to bundle it with this PR for completeness

src/COMPASS/rdr2geo.py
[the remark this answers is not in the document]
LiangJYu: Disregard previous comment. I confused `sas_output_file` with `scratch_path`. rdr2geo modified so output per burst_id now written to it's own directory in scratch.

src/COMPASS/util/runconfig.py
LiangJYu: What should we do if no polarizations are provided and the polarizations in given SAFEs are inconsistent? A mix of [SH, DH] or [SV, DV] SAFEs allowed? Would there be instances near the Arctic where vertical and horizontal polarizations would be mixed?

src/compass/rdr2geo.py
```diff
@@ -0,0 +1,67 @@
...
+    info_channel.log(f"rdr2geo successfully ran in {dt:.3f} seconds")
+
+
+if __name__ == "__main__":
+    '''run rdr2geo from command line'''
+    # load command line args
+    rdr2geo_parser = YamlArgparse()
+
+    # get a runconfig dict from command line args
+    rdr2geo_runconfig = RunConfig.load_from_yaml(rdr2geo_parser.args.run_config_path, 'rdr2geo')
```
vbrancat, line 64: I think we use the wrong `workflow_name` here. Shouldn't this be the name of the radar cslc workflow? Also, the name of the workflow is hard-coded in` RunConfig`

src/compass/rdr2geo.py
```diff
@@ -0,0 +1,67 @@
...
+    proj = isce3.core.make_projection(epsg)
+    ellipsoid = proj.ellipsoid
+
+    for burst in cfg.bursts:
+        # get isce3 objs from burst
+        rdr_grid = burst.as_isce3_radargrid()
+        isce3_orbit = burst.orbit
+
+        # init output directory in scratch
+        output_path = f'{cfg.scratch_path}/{burst.burst_id}'
```
vbrancat, line 34: Is there a way to put together the topo layer and the reference burst? For this workflow, this bundle of rasters will be our product

src/COMPASS/util/runconfig.py
vbrancat: According to the acquisition plan here https://asf.alaska.edu/data-sets/sar-data-sets/sentinel-1/sentinel-1-observation-plans/ there might be regions in the Canadian Arctic where we might have a polarization mix up. However, they will be always belong to two different modes IW and EW (meant for sea ice)

hfattahi: I'm thinking about two users:
1) our SDS:
I don't think if we should force the SDS to specify polarization. In that case then it means that we should any ways lookup the polarization and process only co-pol channel. Note that co-pol is either HH or VV. So far we have committed to process only co-pol channels for OPERA

2) community users:
Many users of coreg-slc are those that will take coreg-slc to interferometry. They usually care about co-pol channel. I think they assume the software looks up the polarization and extracts co-pol. 

NOTE:
In case if in future, OPERA will change the plan to process both polarizations, then we should change the defualt to both pols. In that case community users, should get used to providing their desired polarization.

hfattahi: So to be clear, I think we "polarization" should be "optional" parameter in the run config and the default behavior should be to look up polarization and process co-pol channel.

In case if user asks for a polarization that does not exist, we should throw an exception and exit without falling back to default behavior explained above.

vbrancat:
@gshiroma @yunjunz @jungkyoJung it would be highly appreciated if you can review this PR asap :). Thanks a lot for your help and suggestions.

src/compass/utils/schemas/cslc_s1.yaml
```diff
@@ -34,18 +34,18 @@ runconfig:
         # This section includes parameters to tweak the workflow
         processing:
            # Polarization to process. If not set, process all polarizations for that burst
-           polarization: any(list('HH', 'HV', 'VH', 'VV'), min=1, max=4), str(min=2, max=2), required=False)
-
+           polarization: any(list('HH', 'HV', 'VH', 'VV'), min=1, max=2, required=True)
```
hfattahi, line 37: "polarization" should be optional in my opinion. 
We don't want users (or SDS) be required to choose polarization. By default we should always extract polarization from the filename and process them. I'm hesitant to say if we should process both or a single polarization by default. We have two choices:
1- the default will be to always process co-pol channel
2- add an extra parameter indicating if we want to process co-pol, cross-pol or dual-pol channels. 
Option 2 may make the schema a bit more complicated but has the advantage that the data system can easily switch between co-pol , cross-pol, dual-pol without the need for specifying the exact polarizations. 
If we go with option-2 the existing ```polarization``` parameter should have the highest priority to decide the polarizations but again it should be optional.

src/COMPASS/rdr2geo.py
hfattahi: @LiangJYu I did not see this addressed.

src/COMPASS/util/runconfig.py
hfattahi: @LiangJYu and @vbrancat I don't think if the current behavior of the SAS is completely consistent with what we discussed above. I wrote another comment on the [schema](https://github.com/opera-adt/COMPASS/pull/4/files#r797139128) which may clarify a bit more. Or let me know if more clarification needed or if you disagree.

src/compass/utils/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+        Path to file to be checked
+    """
+    error_channel = journal.error('helpers.check_file_path')
+    if not os.path.isfile(file_path):
+        err_str = f'{file_path} not found'
+        error_channel.log(err_str)
+        raise FileNotFoundError(err_str)
+
+
+def get_file_mode(file_path: str):
```
hfattahi, line 28: We can probably be a bit more clear on this function name. when I saw get_file_mode I thought the function checks if a given file is in read or write mode. But all it does is extracting polarization from filename. How about:
```get_polarization_from_filname``` or ```get_polarization```

File: src/compass/utils/helpers.py [hunk 14]
```diff
@@ -0,0 +1,108 @@
...
+    """Check if given path is a GDAL-compatible file; else raise error
+
+    Parameters
+    ----------
+    dem_path : str
+        File path to DEM for which to check GDAL-compatibility
+    """
+    error_channel = journal.error('helpers.check_dem')
+    try:
+        gdal.Open(dem_path)
```
hfattahi, line 104: We should get used to open in read mode. 
```
gdal.Open(dem_path, gdal.GA_ReadOnly)
```

src/compass/utils/runconfig.py
```diff
@@ -0,0 +1,280 @@
...
+            error_channel.log(err_str)
+            raise ValueError(err_str)
+
+        # Check for SAFE files with non-matching polarization
+        mode_mismatch = any([safe_mode not in cfg_pol_modes
+                             for safe_mode in safe_pol_modes])
+        if mode_mismatch:
+            err_str = "Unexpected safe file polarization mode"
+            error_channel.log(err_str)
+            raise ValueError(err_str)
```
hfattahi, line 82: The logic for polarization should be consistent with what I commented [here](https://github.com/opera-adt/COMPASS/pull/4#discussion_r797139128) and [here](https://github.com/opera-adt/COMPASS/pull/4#discussion_r786960268).

src/compass/utils/runconfig.py
```diff
@@ -0,0 +1,280 @@
...
+    else:
+        # Check if all modes have H
+        if all(['H' in safe_mode for safe_mode in safe_pol_modes]):
+            group_cfg['processing']['polarization'] = ['HH']
+        # Check if all modes have V
+        elif all(['V' in safe_mode for safe_mode in safe_pol_modes]):
+            group_cfg['processing']['polarization'] = ['VV']
```
hfattahi, line 89: I'm not sure if I understand this part! Would you please elaborate what are you trying to do?

LiangJYu: Disregard this. I've updated the polarization logic here c0683a2

src/compass/utils/schemas/cslc_s1.yaml
LiangJYu: I think the changes in c0683a2 should match what you described.

src/compass/utils/helpers.py
```diff
@@ -0,0 +1,123 @@
...
+
+def check_file_path(file_path: str) -> None:
+    """Check if file_path and polarizations exist else raise an error.
+
+    Parameters
+    ----------
+    file_path : str
+        Path to file to be checked
+    polarizations : list[str]
+        Path to file to be checked
```
hfattahi, line 19: Looks like residual from earlier versions that should be removed.

src/compass/utils/helpers.py
```diff
@@ -0,0 +1,123 @@
...
+    file_path : str
+        SAFE file name to parse
+    expected_mode : str
+        (co-pol, cross-pol. dual-pol)
+    '''
+    safe_pol_mode = os.path.basename(file_path).split('_')[-6][2:]
+    if expected_mode != 'co-pol' and safe_pol_mode == 'SV':
+        err_str = f'Cross-pol missing in {safe_file}'
+        error_channel.log(err_str)
+        raise ValueError(err_str)
```
hfattahi, line 45: I think we need a similar check for ```SH``` as well:
```
if (expected_mode != 'co-pol' and safe_pol_mode == 'SV') or (expected_mode != 'co-pol' and safe_pol_mode == 'SH'):
```

src/compass/utils/runconfig.py
```diff
@@ -0,0 +1,246 @@
...
+    _ : list[Sentinel1BurstSlc]
+        List of bursts loaded according to given configuration.
+    '''
+    error_channel = journal.error('runconfig.correlate_burst_to_orbit')
+
+    # dict to store bursts keyed by burst_ids
+    bursts = {}
+
+    # zip pol and IW subswath indices together
+    mode_to_pols = {'co-pol':['VV'], 'cross-pol':['VH'], 'dual-pol':['VV', 'VH']}
```
hfattahi, line 81: why we are missing HH, HV and {HH, HV} ?
I'm sorry probably my many comments about polarization have complicated this part! But I think we are almost there. I very much liked that in the schema you removed HH, VV, VH and HV and replaced with 'co-pol' 'cross-pol' and 'dual-pol'. However, that does not mean that we should completely ignore HH and HV. 
With the current design of the schema for polarization (which I think is the right one), we need the following logic here:

```
mode_to_pols = check_file_polarization_mode(filename) 
 ```
basically we should not hardcode mode_to_pols but rather construct it based on the polarization in filename which would have these options in your check_file_polarization_mode function:
```
if pol == SV:
    mode_to_pols = {'co-pol':['VV']}
elif pol == DV:
    mode_to_pols = {'co-pol':['VV'], 'cross-pol':['VH'], 'dual-pol':['VV', 'VH']}
elif pol == SH:
    mode_to_pols = {'co-pol':['HH']}
else:
    mode_to_pols = {'co-pol':['HH'], 'cross-pol':['HV'], 'dual-pol':['HH', 'HV']}
```

File: src/compass/rdr2geo.py [hunk 20]
```diff
@@ -0,0 +1,103 @@
...
+if __name__ == "__main__":
+    '''run rdr2geo from command line'''
+    # load command line args
+    rdr2geo_parser = YamlArgparse()
+
+    # get a runconfig dict from command line args
+    rdr2geo_runconfig = RunConfig.load_from_yaml(rdr2geo_parser.args.run_config_path, 'rdr2geo')
+
+    # run rdr2geo
+    run(rdr2geo_runconfig)
```
hfattahi, line 103: I think by the time we arrive here ```rdr2geo_runconfig.bursts``` includes all bursts from different polarizations in case if the polarization is dual-pol. Is that right? 
If yes then we will be repeating the same computation of rdr2geo twice (for two pols of same burst) which is really unnecessary and expensive. We need to run rdr2geo only once. 
Similarly later we should run geo2rdr also one for two polarizations of the same burst. Resample will need to be run on each polarization though.

hfattahi (commented):
Thank you @LiangJYu for addressing all my comments. I have few more minor comments in the following. One thing that we did not discuss in the beginning was the fact that in case of dual-pol data we need to run the geometrical computations (rdr2geo and geo2rdr) only once.

src/compass/rdr2geo.py
LiangJYu: Addressed with bc0ec83

src/compass/utils/runconfig.py
LiangJYu: Addressed with 2b53b51

src/compass/rdr2geo.py
hfattahi: Thank you @LiangJYu I think the polarization configuration from schema to evaluations in different steps looks good now. 
However, I don't think if this specific comment here has been addressed. Let's say we have only one burst (one burst ID) in the run config but polarization is set to "dual-pol". In this case "rdr2geo_runconfig.bursts" will be a list of two bursts (same burst ID but two different polarizations). Do you agree?

If correct then by running run(rdr2geo_runconfig) we will be running rdr2geo twice, which is unnecessary because rdr2geo output for the same burst ID is the same for two different polarizations. The efficient way would be to run rdr2geo only one time for the two polarizations of the two bursts.

From an end to end workflow how shall we deal with it. One easy solution can be to check for polarization before calling  RunConfig.load_from_yaml(...) and force only one polarization to load_from_yaml when the function is running for rdr2geo and geo2rdr. 

Any other better solution?

LiangJYu: I will rework. Reverting bc0ec83

LiangJYu: After reviewing bc0ec83, I think it the implementation is mostly correct.

Bursts in a runconfig object are now [stored as](https://github.com/LiangJYu/COMPASS/blob/bc0ec831fa85d32940e5639235354464c8361fa9/src/compass/utils/runconfig.py#L127-L137) `list[list[Sentinel1BurstSlc]]` where each `list[Sentinel1BurstSlc]` have different polarizations and share a common burst ID . This implementation is missing rdr2geo specific logic to check that bursts of the duplicate ID+polarization do not get appended.

Change above impacts `rdr2geo.py` by:
1. [iterating](https://github.com/LiangJYu/COMPASS/blob/bc0ec831fa85d32940e5639235354464c8361fa9/src/compass/rdr2geo.py#L38) over a `list[list[Sentinel1BurstSlc]]` instead of `list[Sentinel1BurstSlc]`.
2. first [writing SLC to ENVI](https://github.com/LiangJYu/COMPASS/blob/bc0ec831fa85d32940e5639235354464c8361fa9/src/compass/rdr2geo.py#L44-L45) for up to two polarizations by iterating over a `list[Sentinel1BurstSlc]`
3. then [running rdr2geo](https://github.com/LiangJYu/COMPASS/blob/bc0ec831fa85d32940e5639235354464c8361fa9/src/compass/rdr2geo.py#L47-L93) once with the last `Sentinel1BurstSlc` from write to ENVI iteration above.

hfattahi:
I could run this PR on the real Sentinel-1 bursts and the results were identical compared with rdr2geo computations from isce2. 

I'm concerned about the lack of unit tests. However, I'm open to merge this PR to pave road for the next  coming PR, open an issue for adding unit tests and work on the unit tests while geo2rdr and resample PR is up.

vbrancat:
@hfattahi I also compared the topo layers from this PR to that from `isce2` and have found that they agree (within numerical precision). For me, this PR is good to go. 
My preference would be to have at least one other unbiased opinion from at least one of the listed reviewers. Again @gshiroma, @yunjunz, @jungkyoJung, this is time sensitive. We would really appreciated your feedbacks on this PR. 

We can have `geo2rdr.py `and `resample.py `packed in one single PR and almost contemporaneously submit a PR for the `rdr2geo` unit test. Hopefully, this sounds like a good plan.

hfattahi:
> @hfattahi I also compared the topo layers from this PR to that from `isce2` and have found that they agree (within numerical precision). For me, this PR is good to go. My preference would be to have at least one other unbiased opinion from at least one of the listed reviewers. Again @gshiroma, @yunjunz, @jungkyoJung, this is time sensitive. We would really appreciated your feedbacks on this PR.
> 
> We can have `geo2rdr.py `and `resample.py `packed in one single PR and almost contemporaneously submit a PR for the `rdr2geo` unit test. Hopefully, this sounds like a good plan.

Sounds good to me.

yunjunz (approved):
LGTM!

src/compass/rdr2geo.py
```diff
@@ -0,0 +1,108 @@
...
+                                            gdal.GDT_Byte)}
+        raster_list = [
+            isce3.io.Raster(f'{output_path}/{fname}.rdr', rdr_grid.width,
+                            rdr_grid.length, 1, dtype, 'ENVI')
+            if enabled else None
+            for fname, (enabled, dtype) in topo_output.items()]
+        x_raster, y_raster, z_raster, layover_shadow_raster = raster_list
+
+        # run rdr2geo
+        rdr2geo_obj.topo(dem_raster, x_raster, y_raster, z_raster,
```
vbrancat, line 86: Do not forget to set the EPSG of the topo.vrt. This will be necessary for the sub-sequent module to properly work

LiangJYu: Output EPSG defaults to `4326` thru [pybind rdr2geo](https://github.com/isce-framework/isce3/blob/73673f757cc30d0d399e9233ecec9008a6be319f/python/extensions/pybind_isce3/geometry/rdr2geo.cpp#L179).

I'm going to add optional `epsg` schema field in `product_path_group`. The default yaml value will be set to 4326.

vbrancat: That was not my experience. The ESPG of the generated output VRT was left unset (i.e. -9999). The following lines:
```
output_vrt = isce3.io.Raster(f'{output_path}/topo.vrt', raster_list)
output_vrt.set_epsg(rdr2geo_obj.epsg_out)
```
solved my issue. I do not think is necessary to expose the EPSG in the yaml.

LiangJYu: `rdr2geo_obj` defaults to `epsg` of 4236. There's no way to change that in the code right now. Is that ok?

vbrancat: Yeah, I am fine with `4326`. My only concern was that the EPSG is not carried over to the VRT (and this will cause `geo2rdr `to Seg fault)

hfattahi (approved):
Thank you @LiangJYu for the nice work. Looks great to me. Thank you for addressing many comments. 
Please open an issue for Unit test before merging.

PR #11, opened 2022-03-10
Geocoded burst workflow

vbrancat:
This PR introduces the geocoded burst workflow for S1-A/B data.
The workflow reads a burst and geocode it on a geogrid that can be entirely specified by the user. 

TO DO:
1. Dump low, main, high band metadata
2. Output the geocoded burst as COG


vbrancat:
@gshiroma @yunjunz @jungkyoJung it would be really appreciated if you can find time to review this PR. This is somewhat critical to subsequent development (stitching, scaling, metadata, logging etc). Thanks a lot.

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # Open valid burst and create output burst. The output burst will
+    # contain 3 bands: Band #1: low-band image; Band #2 main-band image;
+    # Band #3: high-band image.
+    in_ds = gdal.Open(valid_burst_path, gdal.GA_ReadOnly)
+    valid_length = in_ds.RasterYSize
+    valid_width = in_ds.RasterXSize
+    driver = gdal.GetDriverByName('ENVI')
+    out_ds = driver.Create(f'{scratch_path}/{burst.burst_id}_low_main_high',
+                           width, length, 3, gdal.GDT_CFloat32)
```
hfattahi, line 125: why do we need to again save the main-band after split-spectrum?

hfattahi, line 125: why do we need to again save the main-band after split-spectrum?

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,173 @@
...
+    split_spectrum_params = splitspectrum.SplitSpectrum(
+        rg_sample_freq=burst.range_sampling_rate,
+        rg_bandwidth=burst.range_bandwidth,
+        center_frequency=burst.radar_center_frequency,
+        slant_range=rdr_grid.slant_range,
+        freq='A')
+
+    # Save locally the valid part of the burst. Note, invalid data (zeros)
+    # cause issue when computing fft when splitting the range bandwidth
+    valid_burst_path = save_valid_burst(burst, lines_per_block, scratch_path)
```
hfattahi, line 115: Couldn't we do this with VRT? I think it should be easily doable to write a VRT that points to the valid part of the burst.

hfattahi, line 115: Couldn't we do this with VRT? I think it should be easily doable to write a VRT that points to the valid part of the burst.

src/compass/utils/range_split_spectrum.py
hfattahi: we may not need this intermediate file at all. It looks like, few lines below you are reading the data in memory anyways. two options:

1- ignore blocking which should be fine since bursts are small (1500 x 25000 pixels equivalent to ~ 300 MB). 
     burst_data = ds.GetRasterBand(1).ReadAsArray(first_valid_sample, first_valid_line,
                                                        valid_width, valid_length)

2- if you want to keep the blocking (which is preferred), then you should be able to get this to work:
     block_lin = first_valid_line + block * lines_per_block
     burst_data = ds.GetRasterBand(1).ReadAsArray(first_valid_sample, first_valid_line,
                                                        valid_width, block_length)
     you probably need to adjust the logic a bit, but it should work.

hfattahi: we may not need this intermediate file at all. It looks like, few lines below you are reading the data in memory anyways. two options:

1- ignore blocking which should be fine since bursts are small (1500 x 25000 pixels equivalent to ~ 300 MB). 
     burst_data = ds.GetRasterBand(1).ReadAsArray(first_valid_sample, first_valid_line,
                                                        valid_width, valid_length)

2- if you want to keep the blocking (which is preferred), then you should be able to get this to work:
     block_lin = first_valid_line + block * lines_per_block
     burst_data = ds.GetRasterBand(1).ReadAsArray(first_valid_sample, first_valid_line,
                                                        valid_width, block_length)
     you probably need to adjust the logic a bit, but it should work.

hfattahi (changes requested):
Thank you @vbrancat and @LiangJYu for this very exciting PR. Overall looks very good. I have provided some minor comments below which I think can improve the PR before merging. 
I have not been able to run split-spectrum. I think split-spectrum could be a separate PR anyways. Up to you. If you decide to keep split-spectrum in this PR, I will take a more detailed look at that part after initial comments addressed.

src/compass/geo_cslc.py
```diff
@@ -0,0 +1,196 @@
...
+        for burst in bursts:
+            radar_grid = burst.as_isce3_radargrid()
+            native_doppler = burst.doppler.lut2d
+            orbit = burst.orbit
+
+            # Get azimuth polynomial coefficients for this burst
+            az_carrier_poly2d = burst.get_az_carrier_poly()
+
+            # Split the range bandwidth of the burst, if required
+            if cfg.split_spectrum_params.enabled:
```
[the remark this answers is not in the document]
vbrancat: I noticed that I had a bug in the split-spectrum code. The bug should be fixed in the next commits and you should be able to exercise the split-spectrum part of the code. However, it worth clarifying what the ESA folks mean with Hamming window. According to the "Sentinel-1 Level-1 Detailed Algorithm Definition" (https://sentinel.esa.int/documents/247904/1877131/Sentinel-1-Level-1-Detailed-Algorithm-Definition, page 9-14, paragraph 9-7), what is intended with "Hamming window" is instead a family of windows. Eq. 9-21 is what we have implemented in ISCE3 as "raised cosine window". I do not think we should add a new window in the split-spectrum module. We can direct the window type to "raised cosine" whenever we have a Hamming window in the S1 metadata, like is done here: https://github.com/vbrancat/COMPASS/blob/366ef67d240de755f0e514bb96a1e9f00b6d1790/src/compass/utils/range_split_spectrum.py#L88 . Does it sound good?

src/compass/geo_cslc.py
```diff
@@ -0,0 +1,196 @@
...
+                rdr_burst_raster = range_split_spectrum(burst,
+                                                        cfg.split_spectrum_params,
+                                                        scratch_path)
+            else:
+                temp_slc_path = f'{scratch_path}/{burst.burst_id}_temp.tiff'
+                burst.slc_to_file(temp_slc_path, 'GTiff')
```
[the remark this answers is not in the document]
vbrancat: I want to keep the runconfig as simple as possible so I am against of exposing the format as an option. Plus, these temporary files will be ignored by PGE and will be likely be used only for debugging purposes. For runtime performance, Here, some quick numbers when dumping (and geocoding the burst) in GTiff, ENVI and VRT using 32 threads on `aurora.jpl.nasa.gov`:

**geocoding with split-spectrum (3 band raster)**: `GTiff`: 80.16 s, `ENVI`: 80.26, `VRT`: 83.269 s
`geocoding without split-spectrum (1 band raster)`: GTiff`: 52.41 s , `ENVI`:  52.17 s, `VRT`: 51.61 s

When using VRT, the difference in runtime does not seem to be significant; plus, we are saving around 500 MB on disk. Will switch to VRT.

src/compass/geo_cslc.py
```diff
@@ -0,0 +1,196 @@
...
+            # Set geo transformation
+            geotransform = [geo_grid.start_x, geo_grid.spacing_x, 0,
+                            geo_grid.start_y, 0, geo_grid.spacing_y]
+            geo_burst_raster.set_geotransform(geotransform)
+            geo_burst_raster.set_epsg(epsg)
+            del geo_burst_raster
+
+            # Get polygon including valid areas (to be dumped in metadata)
+            filename = f'{output_path}/geo_{burst.burst_id}'
+            poly = get_valid_polygon(filename, np.nan)
```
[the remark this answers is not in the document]
vbrancat: Yes, I do need that for the stitching algorithm. However, I am wondering if we can use the polygon that come directly from Sentinel metadata. That polygon is usually very "accurate". Will hold on this for the moment.

src/compass/utils/geogrid.py
```diff
@@ -0,0 +1,321 @@
...
+    y_start_dict = geo_dict['top_left']['y']
+    x_spacing_dict = geo_dict['x_posting']
+    y_spacing_dict = geo_dict['y_posting']
+    x_end_dict = geo_dict['bottom_right']['x']
+    y_end_dict = geo_dict['bottom_right']['y']
+    x_snap_dict = geo_dict['x_snap']
+    y_snap_dict = geo_dict['y_snap']
+
+    # Check epsg. If None, assign DEM epsg
+    epsg = assign_check_epsg(epsg_dict, dem_raster.get_epsg())
```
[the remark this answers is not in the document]
vbrancat: Ok for the default EPSG from burst center coordinates. I am not sure what do you mean for the bounding box. What are you suggesting to change for the bounding box?

hfattahi:
I have no idea why my comments are duplicated in this review. Sorry about that @vbrancat . I'm not sure why this has happened. Perhaps I had the PR open in two different windows!

hfattahi:
Unfortunately I can not quote the exact comment. This is strange. 
Regarding bounding box above, what I mean is this: If the user has not provided the EPSG and the geogrid, then the SAS should compute it and define the geogrid. The bounding box is in lat-lon, those few lines will convert them to coordinates in epsg. From those coordinates you can get the top-left, bottom-right of the geogrid, I assume

src/compass/utils/range_split_spectrum.py
vbrancat: It needs to be saved because `geocodeSlc` requires, as an input, an `isce3.io.Raster` containing the data to geocode

vbrancat: You are right. We can just dump the burst as it is as a VRT file and read appropriately starting from the valid part of the burst. I have taken this into account in my commit and removed the function saving the valid part of the burst

src/compass/utils/geogrid.py
```diff
@@ -0,0 +1,361 @@
...
+        for burst in common_id_bursts:
+            burst_id = burst.burst_id
+
+            if burst_id in geo_grids:
+                continue
+
+            # Compute Burst epsg if not assigned in runconfig
+            if epsg_dict is None:
+                epsg_default = get_point_epsg(burst.center.y,
+                                              burst.center.x)
```
hfattahi, line 310: You probably need an else for the case when epsg_dict i not None. Otherwise epsg_default is never defined and next line will fail.

src/compass/geo_cslc.py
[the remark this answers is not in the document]
hfattahi: Thanks you @vbrancat for the details. I took a look at the S1 spec you linked to and also I looked at the raised cosine in [isce3](https://github.com/isce-framework/isce3/blob/develop/python/packages/nisar/workflows/focus.py#L68). I'm not sure if the isce3 raised cosine is exactly same as the hamming in the s1 spec. I just tried to create the filter simply based on S1 spec as:

```
W = a-(1-a)*np.cos(np.linspace(0, 2*np.pi, num=nwin, endpoint=True))
```
I also got the filter from isce3. and I also just saw our isce2 implementation [here](https://github.com/isce-framework/isce2/blob/main/components/isceobj/TopsProc/runIon.py#L269). Here is a plot of the three implementation. I don't see that isce3 to be identical to what is in s1 spec. (Note that I shifted the isce2 filter by fftshift). The one labeled as s1 spec is the equation above.


Screen Shot 2022-04-14 at 7 47 58 AM

src/compass/geo_cslc.py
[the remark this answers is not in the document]
hfattahi: I would suggest to be conservative and keep what you have right now. I think we need extensive testing before switching to the polygon within metadata. 
So my suggestion is to keep what you have for now. However, don't you want to store it in the metadata?

src/compass/geo_cslc.py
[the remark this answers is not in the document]
vbrancat: Uhm, strange. I asked @jungkyoJung to check the Hamming window in ISCE3. I will submit a PR to include it, then. What do you mean you don't see any output? If you enabled split-spectrum, your geocoded burst should have 3 bands instead of 1. I tested it yesterday and that was the case.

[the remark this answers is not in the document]
hfattahi: I was looking for three files. LOL. Now I see that you have put them in the same output file with three bands.

src/compass/geo_cslc.py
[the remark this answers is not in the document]
hfattahi: sounds good. Users can always run gdal_translate to write to disk if they want to.

src/compass/utils/geogrid.py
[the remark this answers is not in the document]
hfattahi: Regarding bounding box above, what I mean is this: If the user has not provided the EPSG and the geogrid, then the SAS should compute it and define the geogrid. The bounding box is in lat-lon, those few lines will convert them to coordinates in epsg to define the geogrid. From those coordinates you can get the top-left, bottom-right of the geogrid, I assume.

src/compass/geo_cslc.py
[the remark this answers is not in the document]
oberonia78: Hi @vbrancat @hfattahi , 
The difference between windows in ISCE3 and the others come from how they are implemented. In ISCE3, the raised cosine was implemented as below. 
So, we have to use the adjusted `window_shape` variable to get identical shapes like `window_shape_hamming = 1 - 2* window_shape`. For example, if you used 0.5 for above plot, then we have to use `0` for cosine window so that we can get the identical windows. 
```
def cosine_window(n: int, pedestal: float):
    if not (0.0 <= pedestal <= 1.0):
        raise ValueError(f"Expected pedestal between 0 and 1, got {pedestal}.")
    b = 0.5 * (1 - pedestal)
    t = np.arange(n) - (n - 1) / 2.0
    return (1 - b) + b * np.cos(2 * np.pi * t / (n - 1))
```
comparison_s1_isce3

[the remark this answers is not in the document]
hfattahi: Thanks @oberonia78 for the clarification. So that means we have to adjust the S1 coeffieceient before passing to isce3.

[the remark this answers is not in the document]
oberonia78: Yes. Exactly.

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,118 @@
...
+
+    low_band_freqs = np.array([low_freq_burst, low_freq_burst + low_bw])
+    high_band_freqs = np.array([high_freq_burst - high_bw, high_freq_burst])
+    low_center_freq = low_freq_burst + low_bw / 2
+    high_center_freq = high_freq_burst - high_bw / 2
+
+    # Initialize the split-spectrum parameter object. Note, constructor
+    # requires frequency (A/B) but this is not used anywhere below
+    rdr_grid = burst.as_isce3_radargrid()
+    split_spectrum_params = splitspectrum.SplitSpectrum(
```
hfattahi, line 49: SplitSpectrum is beyond a struct to store some parameters. It is a class that has some methods which does some computations. I suggest to change the name of the instantiated object to better represent the object. My suggestion is:

```suggestion
    split_spectrum_object = splitspectrum.SplitSpectrum(
```

src/compass/utils/geogrid.py
vbrancat: Yes, you are right. I should just compute the default epsg and then the next function `assign_epsg_check` checks if `epsg_dict` is None and if it is assigns the default value computed from the burst center coordinates

src/compass/geo_cslc.py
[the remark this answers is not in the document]
vbrancat: if window_shape is the `window_shape` coefficient coming from S1-A/B metadata, then the coefficient that should be passed to the `isce3` split-spectrum module is `1.0 - window_shape` and `window_type` being raised cosine? Correct @oberonia78 ?

src/compass/geo_cslc.py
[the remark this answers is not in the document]
vbrancat: Sounds good. @LiangJYu will take care to add the polygon inside the metadata file.

src/compass/geo_cslc.py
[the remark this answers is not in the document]
oberonia78: @vbrancat Your understanding is correct except `1.0 - window_shape`. This should be `1.0 - 2 * window_shape` for hamming.

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,118 @@
...
+            window_function=window_type, resampling=False
+        )
+        # Get the high sub-image and corresponding metadata
+        burst_high_data, _ = split_spectrum_params.bandpass_shift_spectrum(
+            slc_raster=burst_data, low_frequency=high_band_freqs[0],
+            high_frequency=high_band_freqs[1],
+            new_center_frequency=high_center_freq,
+            fft_size=valid_width, window_shape=burst.range_window_coefficient,
+            window_function=window_type, resampling=False
+        )
```
hfattahi, line 101: @vbrancat I'm not sure if passing a subset of the SLC (the valid part) to split spectrum is needed. Can you elaborate the reason behind this implementation? Are you concerned about the high-frequencies induced by zeros at the boundaries. How concerning is that really? cutting the signal can be interpreted as shifting the signal which in turn is equivalent to introducing a phase ramp to the spectrum. My concern with the current implementation is that we are simply changing the spectrum. It can be probably avoided but would lead to unnecessarily complicating the workflow. So unless if you have a strong reason behind the current implementation, I would say we should not subset to valid region.

oberonia78: @hfattahi I agree with you. Subsetting SLC could introduce incorrect bandpassing and demodulation as well. Before moving forward, we may need to check what exactly the valid region represents in the S1 data.

vbrancat: @fattahi I was concerned about the high frequency components caused by the zeros in the invalid part. I guess you are suggesting to block process the SLC with its invalid values (zero) and do not worry about them?

src/compass/geo_cslc.py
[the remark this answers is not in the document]
vbrancat: @oberonia78 but the new `window shape` would not be between 0 and 1 anymore, correct? This would cause the split spectrum code to crash. Say, if the window_shape we get from S1-A/B metadata is 0.75, the new window shape is -0.5.

src/compass/utils/range_split_spectrum.py
hfattahi: Correct. No need to subset.

src/compass/utils/geogrid.py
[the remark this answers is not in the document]
vbrancat: I don't think this is necessary, thanks to the last modifications we did to the default `epsg` (thank you!!!). If the user has not provided the `epsg`, we use the `epsg` computed from the burst center coordinates in UTM. This line  https://github.com/vbrancat/COMPASS/blob/54b52f3a198ade378182b92c12c1881cec07d906/src/compass/utils/geogrid.py#L338 will initialize the geogrid based on the default `epsg` (which is in UTM) and updated it based on the info the user has provided. Does it make sense?

vbrancat:
The remaining comments to be addressed are:
1. Include the polygon of valid geocoded data in the metadata file. @LiangJYu will take care of this as probably will require to reorganize the metadata part a bit
2. The modification to the window_shape parameters from S1-A/B metadata prior to use the split-spectrum module.
3. Codacy issues (most of them will go away including the polygon in the metadata)

src/compass/utils/geogrid.py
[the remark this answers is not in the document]
hfattahi: perfect. That's even better.

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,113 @@
...
+            new_center_frequency=low_center_freq,
+            fft_size=width, window_shape=1.0 - burst.range_window_coefficient,
+            window_function=window_type, resampling=False
+        )
+        # Get the high sub-image and corresponding metadata
+        burst_high_data, _ = split_spectrum_obj.bandpass_shift_spectrum(
+            slc_raster=burst_data, low_frequency=high_band_freqs[0],
+            high_frequency=high_band_freqs[1],
+            new_center_frequency=high_center_freq,
+            fft_size=width, window_shape=1.0 - burst.range_window_coefficient,
```
hfattahi, line 97: Not related to this PR really. But let me ask @oberonia78 if the split spectrum can handle fft_size to be any number which equals to or larger than width? in particular does it accept "next power of 2" for efficiency of FFT computation? 
Does split spectrum module expect the user (here this PR) to find the next power of 2 and pass to fft_size here?

oberonia78: @hfattahi Yes. fft_size could be any number if fft_size >= width. Since split_spectrum applies FFT using the given fft_size and computes the frequency range with the given information, fft_size could be any number. 
I thought "next power of 2" could be efficient in computation as you mentioned. But I did not find the big differences between "next power of 2" and image width. It might be because I tested out with UAVSAR images, which is usually smaller than satellite image.

hfattahi: Thanks Jungkyo. Just to be sure, split spectrum expects the user to compute "next power of 2" and pass the corresponding fft size to the split spectrum? is that right?

oberonia78: Nope. The split-spectrum implemented in the ISCE3 does not expect anything. For example, bandpass uses the image width as fft_size.

hfattahi: Ok so it sounds like that splitspectrum does not change the fft_size. So it is user responsibility to introduce the fft_size with correct number. 
@vbrancat my suggestion is that you add a simple function somewhere in your utility to compute next power of 2:

here is one c++ [implementation](https://github.com/isce-framework/isce3/blob/develop/cxx/isce3/signal/Signal.icc#L19) which you can translate to python or we might have it somewhere. An overkill is [here](https://github.com/isce-framework/isce3/blob/develop/cxx/isce3/fft/FFTUtil.icc#L39). The idea is to find the number which is a power of 2 and is larger or equal to width

Once you have that function then compute fft_size from width and then pass the fft_size to split spectrum instead of passing width.

src/compass/geo_cslc.py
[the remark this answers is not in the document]
oberonia78: @vbrancat @hfattahi  Yes. You are right. I'll debug it.

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,137 @@
...
+        Band #2: main band; Band #3: high band
+    '''
+    length, width = burst.shape
+    lines_per_block = cfg_split_spectrum.lines_per_block
+    burst_id_pol = '_'.join([burst.burst_id, burst.polarization])
+
+    # In ISCE3, we can use raised cosine to implement S1-A/B Hamming
+    window_type = burst.range_window_type
+    window_type = 'Cosine' if window_type.casefold() == 'hamming' else window_type
+
```
oberonia78, line 54: I think we need to add one line here to adjust the window coefficient.
```
window_shape = 2 * burst.range_window_coefficient - 1 if window_type.casefold() == 'hamming' else burst.range_window_coefficient 
```

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,137 @@
...
+        # Read a block of valid burst data
+        burst_data = in_ds.GetRasterBand(1).ReadAsArray(0, line_start,
+                                                        width, block_length)
+        # Get the low band sub-image and corresponding metadata
+        burst_low_data, _ = split_spectrum_obj.bandpass_shift_spectrum(
+            slc_raster=burst_data, low_frequency=low_band_freqs[0],
+            high_frequency=low_band_freqs[1],
+            new_center_frequency=low_center_freq,
+            fft_size=find_next_power(width),
+            window_shape=1.0 - burst.range_window_coefficient,
```
oberonia78, line 112:
```suggestion
            window_shape=window_shape
```

src/compass/utils/range_split_spectrum.py
```diff
@@ -0,0 +1,137 @@
...
+            window_function=window_type, resampling=False
+        )
+        # Get the high sub-image and corresponding metadata
+        burst_high_data, _ = split_spectrum_obj.bandpass_shift_spectrum(
+            slc_raster=burst_data, low_frequency=high_band_freqs[0],
+            high_frequency=high_band_freqs[1],
+            new_center_frequency=high_center_freq,
+            fft_size=find_next_power(width),
+            window_shape=1.0 - burst.range_window_coefficient,
```
oberonia78, line 121:
```suggestion
            window_shape=window_shape,
```

src/compass/utils/range_split_spectrum.py
oberonia78, line 54:
```suggestion
window_shape = 2 * burst.range_window_coefficient - 1 if window_type.casefold() == 'hamming' else burst.range_window_coefficient 
```

src/compass/geo_cslc.py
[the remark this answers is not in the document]
hfattahi: Ok to resolve this comment based on our discussion on isce3 repo, it looks like we should pass  "window_shape = 2 * burst.range_window_coefficient - 1", assuming S1 spec enforces the coefficient to be between 0.5 to 1.0 
@oberonia78 for completeness, would you please confirm and point to S1 documentation that the window coefficient is between 0.5 to 1.0 ?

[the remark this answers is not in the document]
hfattahi: Thank you.

src/compass/utils/schemas/geo_cslc_s1.yaml
```diff
@@ -0,0 +1,104 @@
...
+        pge_name_group:
+           pge_name: enum('CSLC_S1_PGE')
+
+        input_file_group:
+           # Required. List of SAFE files (min=1)
+           safe_file_path: list(str(), min=1)
+           # Required. List of orbit (EOF) files
+           orbit_file_path: list(str(), min=1)
+           # List of (unique) burst ID to process
+           burst_id: list(str(), min=1, required=True)
```
hfattahi, line 14: @vbrancat my understanding is that geo_cslc.py processes only one burst (which can have one or more polarizations). Is this correct? If yes, shall we change list to str?

vbrancat:
For some reason, I am not able to comment on previous comments. I have noticed that in the split spectrum processing, if I set the width to the next power of two of the burst width, it causes the following issue in the geocoded burst:
Screen Shot 2022-04-18 at 3 56 52 PM

Is the next power of two required to speed up the fft? I have not noticed that this makes a significant difference

hfattahi:
> For some reason, I am not able to comment on previous comments. I have noticed that in the split spectrum processing, if I set the width to the next power of two of the burst width, it causes the following issue in the geocoded burst: Screen Shot 2022-04-18 at 3 56 52 PM
> 
> Is the next power of two required to speed up the fft? I have not noticed that this makes a significant difference

I suggest to not change the interface here. Open an issue in isce3 with this observation. In the meantime we can return ```width``` in find_next_power function until isce3 issue resolves.

src/compass/utils/schemas/geo_cslc_s1.yaml
vbrancat: That is correct. The current baseline is to process burst by burst. We can enforce the orbit file and the `burst_id` to be `str` instead of list of strings.

hfattahi:
@LiangJYu not related to this PR, However [this logic](https://github.com/opera-adt/COMPASS/pull/11/files#diff-2ffcbcbbe9301bfe503db2eedc97bd09159a8c2eb06fc50afd3fe705e4d7116fL115
) does not allow processing dual-pol data. for dual-pol we have pols = ["VV", "VH"] and i_subswaths = [1,2,3] 

we expect zip_list  to contain 

VV 1, VV 2 , VV 3, VH 1, VH 2, VH 3 

which is not what I observed. 

What we really need in this case can be obtained with:

pairs = itertools.product(pols, i_subswaths)
result = [pair for pair in pairs if pair[0] != pair[1]]

src/compass/geo_cslc.py
```diff
@@ -0,0 +1,123 @@
...
+    runconfig = GeoRunConfig.load_from_yaml(geo_parser.run_config_path,
+                                            'geo_cslc_s1')
+
+    # Run geocode burst workflow
+    run(runconfig)
+
+    # Save burst metadata and runconfig parameters
+    json_path = f'{runconfig.product_path}/metadata.json'
+    with open(json_path, 'w') as f_json:
+        runconfig.to_file(f_json, 'json')
```
hfattahi, line 123: It would be great to name the metadata file consistent with the geocoded product. I suggest to move these few lines to run and use f'{output_path}/geo_{burst_id}_{pol}.json' for the name of the json file.

src/compass/utils/schemas/geo_cslc_s1.yaml
hfattahi: I think it is ok to keep safe_file and orbit_file as "list". But burst_id should be "str" because the workflow handles one burst_id. 
The reason to keep afe_file and orbit_file as "list" would be the rare situation that multiple zip files are given and user is not sure which one has the bust_id.

src/compass/utils/geo_runconfig.py
```diff
@@ -0,0 +1,127 @@
...
+            Format of output
+        add_burst_boundary: bool
+            If true add burst boundary string to each burst entry in dict.
+            Reads geocoded burst rasters; only viable after running geo_cslc.
+        '''
+        self_as_dict = self.as_dict()
+
+        if add_burst_boundary:
+            for b_dict in self_as_dict['bursts']:
+                add_poly_to_dict(some_dir, b_dict)
```
hfattahi, line 119: what is some_dir?

src/compass/geo_cslc.py
hfattahi: The polygon representing the valid geocoded burst boundary not written to metadat file by default and here is not enabled. Any reason why not writing them?

src/compass/utils/runconfig.py
```diff
@@ -285,3 +333,27 @@ def gpu_enabled(self):
...
+
+    def as_dict(self):
+        # convert to dict first then dump to yaml
+        self_as_dict = {}
+        for key, val in self.__dict__.items():
+            if key == 'groups':
+                val = unwrap_to_dict(val)
+            elif key == 'bursts':
+                # just date in datetime obj as string
+                date_str = lambda b : str(b.sensing_start.date()).split()[0]
```
hfattahi, line 345:
```suggestion
                date_str = lambda b : b.sensing_start.date().strftime("%Y%m%d")
```
I was just trying to load the metadat "metadata.json" and did a quick experiment to convert the dict to namespace using 

```
import json
from compass.utils.wrap_namespace import wrap_namespace

fid = open("metadata.json","r")
m = json.load(fid)
s = wrap_namespace(m)
```
and then I can say:

```
s.bursts
```
but it fails for 

```
 s.bursts.t173_iw2_b863_2017-05-10_VH
  File "<ipython-input-6-e3d644dcbbbd>", line 1
    s.bursts.t173_iw2_b863_2017-05-10_VH
                                 ^
SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers
```
And that is because Python packages and modules can not use hyphens, only underscores. 

If you go with my suggestion above, then we can say:

```
In [11]: s.bursts.t173_iw2_b863_20170510_VV.sensing_start
Out[11]: '2017-05-10 13:43:38.353501'```

which is cool! i.e., we can access the metadata as the members of a python object.

File: src/compass/geo_cslc.py [hunk 18]
```diff
@@ -0,0 +1,123 @@
...
+    ellipsoid = proj.ellipsoid
+    image_grid_doppler = isce3.core.LUT2d()
+    threshold = cfg.geo2rdr_params.threshold
+    iters = cfg.geo2rdr_params.numiter
+    blocksize = cfg.geo2rdr_params.lines_per_block
+    dem_margin = cfg.geocoding_params.dem_margin
+    flatten = cfg.geocoding_params.flatten
+
+    for burst in cfg.bursts:
+        date_str = str(burst.sensing_start.date())
```
hfattahi, line 44:
```suggestion
        date_str = burst.sensing_start.date().strftime("%Y%m%d")
```

LiangJYu:
> @LiangJYu not related to this PR, However [this logic](https://github.com/opera-adt/COMPASS/pull/11/files#diff-2ffcbcbbe9301bfe503db2eedc97bd09159a8c2eb06fc50afd3fe705e4d7116fL115) does not allow processing dual-pol data. for dual-pol we have pols = ["VV", "VH"] and i_subswaths = [1,2,3]
> 
> we expect zip_list to contain
> 
> VV 1, VV 2 , VV 3, VH 1, VH 2, VH 3
> 
> which is not what I observed.
> 
> What we really need in this case can be obtained with:
> 
> pairs = itertools.product(pols, i_subswaths) result = [pair for pair in pairs if pair[0] != pair[1]]

@hfattahi Thanks for finding this. I will switch `itertools` with list comprehension.
```bash
In [13]: pols = ['HH']

In [14]: for pol, i in [(pol, i) for pol in pols for i in i_subswaths]: print(pol, i)
HH 1
HH 2
HH 3

In [15]: pols = ['HH', 'HV']

In [16]: for pol, i in [(pol, i) for pol in pols for i in i_subswaths]: print(pol, i)
HH 1
HH 2
HH 3
HV 1
HV 2
HV 3
```

src/compass/geo_cslc.py
LiangJYu: > The polygon representing the valid geocoded burst boundary not written to metadat file by default and here is not enabled. Any reason why not writing them?

I forgot to enable polygon write; will enable it here.

LiangJYu: > It would be great to name the metadata file consistent with the geocoded product. I suggest to move these few lines to run and use f'{output_path}/geo_{burst_id}_{pol}.json' for the name of the json file.

@hfattahi Currently all the burst metadata is stored in a single file along with runconfig attributes. Sounds like you would like the burst metadata stored separately from runconfig attributes?

src/compass/utils/runconfig.py
hfattahi: For completeness I think we should only write the metadata that are relevent to the output geocoded burst and avoid writing those that are specifically pointing to the input burst. After moving this part to within ```run```, let's create a new dict let's say called ```gslc_dict``` which has the following keys:

dict_keys(['sensing_start', 'radar_center_frequency', 'wavelength', 'azimuth_steer_rate', 'azimuth_time_interval', 'slant_range_time', 'starting_range', 'range_sampling_rate', 'range_pixel_spacing', 'azimuth_fm_rate', 'doppler', 'range_bandwidth', 'polarization', 'burst_id', 'platform_id', 'center', 'border', 'orbit', 'tiff_path', 'i_burst', 'range_window_type', 'range_window_coefficient'])

which we can borrow from ```runconfig['bursts'][burst.burst_id]```

Let's add sensing_stop (you need to take care of the datetime object)
gslc_dict['sensing_stop'] =  gslc_dict['sensing_start'] + radargrid.length *  azimuth_time_interval

and then let's add the geogrid
gslc_dict['geogrid'] = runconfig['geogrids'][burst.burst_id]  

and add run config
gslc_dict["runconfig"] = runconfig['groups']

add the nodata value:
gslc_dict['nodata'] = NO_DATA_VALUE

we should read this with the reader
gslc_dict['input_data_ipf_version'] =  

empty for now but would be nice to keep track of it. 
gslc_dict['compass-isce3-version'] = ""



And then let's write gslc_dict to file (json, yaml). That way when we load it back as:
```
fid = open("metadata.json","r")
m = json.load(fid)
s = wrap_namespace(m)
```
this time we can easily access the metadata at the top level of the object as for example:

```
s.sensing_start
```
in opposed to what we had above as

```
s.bursts.t173_iw2_b863_20170510_VV.sensing_start
```
You may make sure that s.polarization represents all polarizations in the output geocoded burst. 

This should be most of the light metadata. @vbrancat @yunjunz  can you think of other light metadata that may be useful to add.  I can imagine that we will add bunch of metadata about correction when we get there.

hfattahi: @LiangJYu and @vbrancat sorry this is becoming a bit long comment. Happy to talk offline if needed. What do you think about this idea to define a class for geocoded burst slc which has reader and writer methods. Here is a very general idea to build on:

```
class GSLC:
    def __init__(self):
    self.sensing_start = None
    self.sensing_stop = None
    self.polarization = None
    self.orbit = None
    # ....

    def as_dict(self):
        return gslc_dict

    def sensing_start(self, s):
        self.sensing_start = s

    def sensing_stop(self, s):
        self.sensing_stop = s

    # and setters for other parameters ...

    def load_from_dict(self, input_dict):
        # We can assume input dict is the runconfig Then for max clarity we could
        # go one by one and set all data memebers

    def write_metadata(self, metadata_file):
        self_as_dict = self.as_dict()
        with open(metadata_file, 'w') as f_json:
            json.dump(self_as_dict, f_json)

    def load_metadata(self, metadata_file):
        fid = open("metadata.json","r")
        meta_dict = json.load(fid)
        self = wrap_namespace(meta_dict)
```
I think this will significantly increase the clarity and will make the maintenance of the product and its metadata much easier. It is much more clear to see each of the metadata (class data members) and how they are set before writing them to file. Also with a simple method we should be able to reconstruct the object from file.

vbrancat: I like the idea of the class and it can simplify a lot the product reader and writer (which is on our TO DO list anyway). Plus, the class can be used in other parts of the code (e.g. stitching). @LiangJYu will take care of the metadata part

yunjunz:
This PR needs to be updated against the latest main branch, as there are some conflicts flagged by github.

LiangJYu:
> This PR needs to be updated against the latest main branch, as there are some conflicts flagged by github.

I have already merged with main. I'll push it along with the metadata fixes later this afternoon (hopefully).

LiangJYu:
[9104b5e](https://github.com/opera-adt/COMPASS/pull/11/commits/9104b5ee60222df5ed994bf314daa869253e3213) addresses most comments and can be run for testing. A bit of polishing of metadata is left to be done.

src/compass/utils/runconfig.py
yunjunz: For the metadata, the following could also be useful: 'length', 'width', 'earth_radius', 'satellite_height', 'orbit_direction', 'antenna_side', 'pulse_repetition_frequency' and the EPSG code.

hfattahi: @yunjunz length, width and epsg are must to have and included in geogrid mentioned above. what do you mean by antenna_side? Do you mean look_direction? If yes I prefer look_direction compared to antenna_side.

The orbit is included which is the satellite XYZ and velocity. I'm not sure if we should provide satellite_height and earth_radius. satellite_height if needed , they can be derived from the orbit info. One can convert XYZ to llh, look up DEM height at ll and subtract from h of llh to get to satellite_height. 

It would be good to have orbit_direction in the metadata.

yunjunz: @hfattahi yes, for the antenna_side, I meant "look_direction", that sounds good to me too. 

The calculation of satellite_height and the used earth_radius is not very end user friendly, I think it will be useful to have them unless this causes confusion on our side.

LiangJYu:
@hfattahi @yunjunz Commit d9b3071 should address most comments and concerns on metadata with IMO minimal boilerplate.

src/compass/utils/runconfig.py
LiangJYu: @hfattahi @yunjunz `look direction` easier to add to s1_reader/s1_burst. Will simply returning `right` be sufficient?

`orbit_direction` will also be easier to add to s1_reader/s1_burst.

hfattahi: I think look direction returning from s1 reader is fine. And yes look direction for S1 is fixed to "right". 
And yes orbit direction can come with the reader.

LiangJYu:
@06612d1 eliminates unwanted attributes based on the top of this [comment](https://github.com/opera-adt/COMPASS/pull/11#discussion_r853632510). Removing `doppler.lut2d` would significantly reduce the volume of metadata since `doppler.lut2d` has shape on the order (2, 2e4).

vbrancat:
I had a quick look to the JSON metadata file produced by the latest code modifications. I have a couple of considerations:

- Why do we allocate the polygon from the burst metadata and the one we compute (which surely includes valid pixel data). Do we really need both?
- `range_bandwidth` is ambiguous. I assume this refers to the main band of the burst. We also include a low and side band rasters. We should probably have 3 entries something like `low_band_range_bandwidth`, `main_band_range_bandwidth`, `high_band_range_bandwidth`.
- There is an inconsistency in the selected coordinate system. For example, we allocate the burst center coordinates (coming from the metadata) in lat/lon but we provide the geogrid bbox in UTM. We should move everything to UTM. 
- We do not have info on the shape of the geocoded burst. Wondering if we would like to have X-Y Coordinate vectors
- We should better reorganize the metadata structure. For example, it might be worth put together metadata with same broad characteristics e.g., sensor parameter, radar geometry parameters, processing parameters.

LiangJYu:
> I had a quick look to the JSON metadata file produced by the latest code modifications. I have a couple of considerations:
> 
> Why do we allocate the polygon from the burst metadata and the one we compute (which surely includes valid pixel data). Do we really need both?

Discard the one from SAFE annotation XML and keep the one computed with shapely?

> `range_bandwidth` is ambiguous. I assume this refers to the main band of the burst. We also include a low and side band rasters. We should probably have 3 entries something like `low_band_range_bandwidth`, `main_band_range_bandwidth`, `high_band_range_bandwidth`.

Use current `range_bandwidth` from SAFE annotation XML as `main_band` and only populate `low` and `high` when provided?

> There is an inconsistency in the selected coordinate system. For example, we allocate the burst center coordinates (coming from the metadata) in lat/lon but we provide the geogrid bbox in UTM. We should move everything to UTM.

Current center value is computed as the average of border from the SAFE. Switch with average of the shapely boundary and thus keeping units consistent?

> We do not have info on the shape of the geocoded burst. Wondering if we would like to have X-Y Coordinate vectors

Do the contents of [geogrid](https://github.com/opera-adt/COMPASS/blob/55143c625e77a040e8d25818b4809ac5ecf52380/src/compass/utils/geo_cslc_metadata.py#L158-L162) suffice?

> We should better reorganize the metadata structure. For example, it might be worth put together metadata with same broad characteristics e.g., sensor parameter, radar geometry parameters, processing parameters.

I would prefer reorganizing the GCSLC metadata after burst SLC attribute reorganizing because of the high degree overlap. The burst SLC class is also due for additional [refactor](https://github.com/opera-adt/s1-reader/discussions/16).

hfattahi:
All wonderful points @vbrancat .

> I had a quick look to the JSON metadata file produced by the latest code modifications. I have a couple of considerations:
> 
> * Why do we allocate the polygon from the burst metadata and the one we compute (which surely includes valid pixel data). Do we really need both?

I agree we need the valid region polygon only. And we probably need to clarify the name of the attribute to represent the valid region?

> * `range_bandwidth` is ambiguous. I assume this refers to the main band of the burst. We also include a low and side band rasters. We should probably have 3 entries something like `low_band_range_bandwidth`, `main_band_range_bandwidth`, `high_band_range_bandwidth`.

I'm not sure about this. There is no side-band for sentinel-1. So why we should use main-band? 
Regardless of that detail, we have two choices here to deal with split spectrum:
1- have a dynamic spec that allows for the metadata of the SLC and when there is split spectrum, allow for adding their metadata added to the same file.

2- treat each of the sub-bands as one SLC. So the spec is static and always have one frequency and one bandwidth representing that specific SLC. So if we happen to split the spectrum and end up with three SLCs then we will have 3 metadata files. But no matter where they came from, they have their own range-bandwidth and center-frequecny.  

I personally prefer the second option. It's much cleaner.  

> * There is an inconsistency in the selected coordinate system. For example, we allocate the burst center coordinates (coming from the metadata) in lat/lon but we provide the geogrid bbox in UTM. We should move everything to UTM.

I agree!

> * We do not have info on the shape of the geocoded burst. Wondering if we would like to have X-Y Coordinate vectors
I don't think if we need X-Y coord vectors. The reason we had that in NISAR spec was to conform to CF convention. Otherwise even over there, it was not needed. The grid is regular and X and Y can be computed from the geogrid which is actually stored in the metadata. I see the following in the stored json:

```
"geogrid": {
        "start_x": 598450.0,
        "start_y": 3966600.0,
        "spacing_x": 5.0,
        "spacing_y": -10.0,
        "length": 3485,
        "width": 18580,
        "epsg": 32611
``` 
But perhaps it could be stored outside the run config section?!

> * We should better reorganize the metadata structure. For example, it might be worth put together metadata with same broad characteristics e.g., sensor parameter, radar geometry parameters, processing parameters.

I have mixed feelings on this one. But I'm open for discussion. On one hand I can see the possible value and clarity on the other hand I'm afraid the boundary may not be very clear between the different groups. For example I would argue that range bandwidth belongs to all those three categories you mentioned. So in which one do you want to put? What will be really the gain if they are categorized? 

Does that re-organization affect how to access the metadata? For example right now you can say:

```
m = GeoCslcMetadata.from_file("geo_t173_iw2_b863_VV.json", "json")
```
And then you have directly access to all metadata at the same level, e.g. see the screenshot:



Screen Shot 2022-04-28 at 6 10 15 PM

So do we prefer to say ```m.sensing_start``` or ```m.sensor.sensing_start``` ?

vbrancat:
I think at this stage all the comments have been addressed. The PR is again ready to be reviewed @hfattahi @oberonia78 @shiroma @yunjunz

hfattahi (approved):
Thank you very much @vbrancat and @LiangJYu for your patience and addressing my comments. This is a very exciting addition. Congratulations for your nice work. I don't have any more comments and I think this PR can be merged. There will be improvements and additions/modifications to the metadata etc which is beyond the scope of this PR. Thanks again for your nice work.

src/compass/utils/defaults/geo_cslc_s1.yaml
```diff
@@ -0,0 +1,68 @@
...
+  groups:
+      pge_name_group:
+          pge_name: CSLC_S1_PGE
+
+      input_file_group:
+          # Required. List of SAFE files (min=1)
+          safe_file_path:
+          # Required. List of orbit (EOF) files (min=1)
+          orbit_file_path:
+          # Required. List of unique burst IDs to process
```
yunjunz, line 13:
```suggestion
          # Required. The unique burst ID to process
```

yunjunz:
The PR runs nicely on my end. Testings with the two SLCs generate the file structure below. I have two suggestions regarding the file names:

1. Could we add the `.slc` file extension to the SLC file?
2. Since the burst_id is in the folder name, should we remove it from all the file names, to be more consistent with the RSLC workflow?

```bash
(opera) kamb:~/data-marmot/OPERA/compass_gslc>$ tree gslc/
gslc/
└── t71_iw1_b848
    ├── 20200511
    │   ├── geo_t71_iw1_b848_VV
    │   ├── geo_t71_iw1_b848_VV.hdr
    │   ├── geo_t71_iw1_b848_VV.json
    │   └── t71_iw1_b848_VV_temp.vrt
    └── 20200517
        ├── geo_t71_iw1_b848_VV
        ├── geo_t71_iw1_b848_VV.hdr
        ├── geo_t71_iw1_b848_VV.json
        └── t71_iw1_b848_VV_temp.vrt

3 directories, 8 files
```

src/compass/utils/schemas/cslc_s1.yaml
```diff
@@ -11,7 +11,7 @@ runconfig:
            # Required. List of orbit (EOF) files
            orbit_file_path: list(str(), min=1)
            # List of (unique) burst IDs to process
-           burst_id: list(str(), min=1, required=True)
+           burst_id: str(required=True)
```
yunjunz, line 14: This changes the behavior of the RSLC workflow. Is it intentional? I am not sure if I missed something in the conversion.

src/compass/utils/schemas/geo_cslc_s1.yaml
```diff
@@ -0,0 +1,104 @@
...
+    groups:
+        pge_name_group:
+           pge_name: enum('CSLC_S1_PGE')
+
+        input_file_group:
+           # Required. List of SAFE files (min=1)
+           safe_file_path: list(str(), min=1)
+           # Required. List of orbit (EOF) files
+           orbit_file_path: list(str(), min=1)
+           # List of (unique) burst ID to process
```
yunjunz, line 13:
```suggestion
           # (unique) burst ID to process
```

src/compass/utils/defaults/geo_cslc_s1.yaml
```diff
@@ -0,0 +1,68 @@
...
+                  y:
+          geo2rdr:
+              lines_per_block: 1000
+              threshold: 1.0e-8
+              numiter: 25
+          range_split_spectrum:
+              enabled: False
+              lines_per_block: 1000
+              low_band_bandwidth:
+              high_band_bandwidth:
```
yunjunz, line 59: Could you add an example/default value for these two parameters?

hfattahi:
> The PR runs nicely on my end. Testings with the two SLCs generate the file structure below. I have two suggestions regarding the file names:
> 
> 1. Could we add the `.slc` file extension to the SLC file?
> 2. Since the burst_id is in the folder name, should we remove it from all the file names, to be more consistent with the RSLC workflow?
> 
> ```shell
> (opera) kamb:~/data-marmot/OPERA/compass_gslc>$ tree gslc/
> gslc/
> └── t71_iw1_b848
>     ├── 20200511
>     │   ├── geo_t71_iw1_b848_VV
>     │   ├── geo_t71_iw1_b848_VV.hdr
>     │   ├── geo_t71_iw1_b848_VV.json
>     │   └── t71_iw1_b848_VV_temp.vrt
>     └── 20200517
>         ├── geo_t71_iw1_b848_VV
>         ├── geo_t71_iw1_b848_VV.hdr
>         ├── geo_t71_iw1_b848_VV.json
>         └── t71_iw1_b848_VV_temp.vrt
> 
> 3 directories, 8 files
> ```

We don't need to be consistent with the radar coreg slc. But let's make the right choice here. 
adding ".slc" sounds ok. But I'm not sure if we should remove the burst_id
SDS will rename these filenames anyways. However for our own use, I think it is useful to keep the burst ID in the name. I like that it also has polarization. I suggest to remove "geo" and add date to the name if possible. I think that will make the name unique. So one can copy all of them in one directory for down stream workflows and things should still work.

yunjunz:
With the split-spectrum turned on, the output file is as below. Both `geo_{burst_id}_VV` and `{burst_id}_VV_low_main_high` have 3 bands. May I ask what is the `geo_{burst_id}` file for?

```bash
(opera) kamb:~/data-marmot/OPERA/compass_gslc/gslc/t71_iw1_b848/20200511>$ ll
total 1.6G
drwxr-xr-x 2 zyunjun    8 May  5 18:08 ./
drwxr-xr-x 4 zyunjun    4 May  5 17:23 ../
-rw-r--r-- 1 zyunjun 1.8G May  5 18:04 geo_t71_iw1_b848_VV
-rw-r--r-- 1 zyunjun  549 May  5 18:04 geo_t71_iw1_b848_VV.hdr
-rw-r--r-- 1 zyunjun 7.9K May  5 18:04 geo_t71_iw1_b848_VV.json
-rw-r--r-- 1 zyunjun 735M May  5 18:01 t71_iw1_b848_VV_low_main_high
-rw-r--r-- 1 zyunjun  138 May  5 18:01 t71_iw1_b848_VV_low_main_high.hdr
-rw-r--r-- 1 zyunjun  828 May  5 18:01 t71_iw1_b848_VV_temp.vrt
```

vbrancat:
@hfattahi I will implement the modification and push the changes to the code. 
@yunjunz there is an imprecision in how you have set the runconfig.  The `product_path_group` has 2 entries: `product_path` where the actual geocoded burst is stored and `scratch_dir` where temporary files are stored. Looking at your directory structure, it seems that you have set the above to point at the same location. When the split spectrum is enabled we generate a temporary file in radar coordinates containing 3 bands (low, main and high band) that is your `t71_iw1_b848_VV_low_main_high`. This gets geocoded to `geo_t71_iw1_b848_VV` which also has 3 bands. In a separate PR, we will make sure that when split-spectrum is activated we will generate 3 geocoded bursts with one band each. This solution is much cleaner from a data system perspective.

src/compass/utils/schemas/cslc_s1.yaml
vbrancat: Yes, this is intentional. Both the workflows will process one burst at the time. I modified the comments to be consistent.

vbrancat:
@hfattahi @yunjunz This is the new directory/filename structure for the output product:

```
product_dir_geo/
└── t71_iw3_b848
    └── 20200517
        ├── t71_iw3_b848_20200517_VV.json
        └── t71_iw3_b848_20200517_VV.slc

2 directories, 2 files
```
Let me know if you want me to further simplify it. Plus, may be we should rename `geo_cslc.py` as `s1_geo_cslc.py` for consistency?

yunjunz:
Thank you @vbrancat for the update, it looks great, I don't have further suggestions on the file structure.

As for the code renaming, either doing it here or in another PR sounds good to me.

src/compass/utils/runconfig.py
```diff
@@ -199,21 +263,18 @@ def get_ref_radar_grid_info(ref_path, burst_ids):
...
+        raise FileNotFoundError(f'Reference radar grid not found for {burst_id}')
 
-        if len(b_id_rdr_grid_files) > 1:
-            raise FileExistsError(f'More than one reference radar grid found for {burst_id}')
+    if len(b_id_rdr_grid_files) > 1:
+        raise FileExistsError(f'More than one reference radar grid found for {burst_id}')
 
-        ref_rdr_path = os.path.dirname(b_id_rdr_grid_files[0])
-        ref_rdr_grid = file_to_rdr_grid(b_id_rdr_grid_files[0])
-        ref_rdr_grids[burst_id] = ReferenceRadarInfo(ref_rdr_path, ref_rdr_grid)
```
yunjunz, line 214: Testing on the RSLC workflow for a secondary date gives the following error. It seems to be related to the changes here.

```bash
(opera) kamb:~/data-marmot/OPERA/compass_rslc>$ python -m compass.s1_cslc cslc_s1_20200517.yaml 
journal: Starting geo2rdr
  File "/home/zyunjun/tools/miniconda3/envs/opera/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/zyunjun/tools/miniconda3/envs/opera/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/zyunjun/tools/COMPASS/src/compass/s1_cslc.py", line 25, in <module>
    run(runconfig)
  File "/home/zyunjun/tools/COMPASS/src/compass/s1_cslc.py", line 12, in run
    s1_geo2rdr.run(cfg)
  File "/home/zyunjun/tools/COMPASS/src/compass/s1_geo2rdr.py", line 70, in run
    burst_path = cfg.reference_radar_info[burst_id].path
TypeError: __getitem__(): incompatible function arguments. The following argument types are supported:
    1. (self: pybind_isce3.product.RadarGridParameters, arg0: tuple) -> pybind_isce3.product.RadarGridParameters

Invoked with: <pybind_isce3.product.RadarGridParameters object at 0x7f350740a970>, 't71_iw1_b848'
```

vbrancat: I will leave this to @LiangJYu who has followed the radar workflow more closely.

yunjunz:
My last two comments would be: 1) the example/default value on the low/high range bandwidth in the YAML file and 2) the codacy flagged issues.

src/compass/utils/runconfig.py
LiangJYu: Fixed with 649cbbe.

src/compass/utils/runconfig.py
```diff
@@ -246,65 +307,20 @@ def load_from_yaml(cls, yaml_path: str, workflow_name: str) -> RunConfig:
...
-        validate_group_dict(default_cfg['runconfig']['groups'])
+        cfg = load_validate_yaml(yaml_path, workflow_name)
 
         # Convert runconfig dict to SimpleNamespace
-        sns = wrap_namespace(default_cfg['runconfig']['groups'])
+        sns = wrap_namespace(cfg['runconfig']['groups'])
 
         bursts = runconfig_to_bursts(sns)
 
         # Load reference grids if not reference run i.e. not running rdr2geo
```
yunjunz, line 317:
```suggestion
        # Load reference grids if not reference run i.e. not running rdr2geo
        ref_rdr_grid_info = None
```

This seems to fix the UnboundLocalError of `ref_rdr_grid_info` while testing the reference rslc

src/compass/utils/runconfig.py
yunjunz: Thank you @LiangJYu for the quick fix.

src/compass/utils/defaults/geo_cslc_s1.yaml
vbrancat: Thanks. During my tests, I set the low- and high-band bandwidths based on Curen paper https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=8706258. I am not sure if this is optimal, only more testing can tell. Once this is done, can you approve the PR? Codacy has been address by Liang. Merci

PR #14, opened 2022-04-04
Stitching geocoded bursts for stack processing

vbrancat:
This PR provides a script to stitch S1-A/B bursts for stack processing i.e., the stitched bursts forming the stitched SLCs have all the same shape and can be directly interfered. To efficiently handle the list of burst IDs to stitch, the PR makes use of `pandas` which would be a new (but light) dependency for `COMPASS`.

The algorithm implements the following steps:

1. If a list of burst IDs to stitch is not provided, it identifies a common list of bursts among all the different dates and use those burst ID for stitching. Otherwise, only the provided list of burst IDs is used for stitching.

2. For each unique burst ID, it identifies the common burst boundary (on the ground) among the different dates and use the common burst boundary to cut the bursts at different dates. This is implemented by saving a `shapely.geometry.Polygon` into an `ESRI Shapefile`. This can be used by `gdal.Warp` to cut the different bursts using the `cutline` feature

3. All the cut bursts for a same date are then stitched together to form the output SLCs. All the formed SLCs have the same shape as they have been formed by using the commonly identified bursts.

**Assumptions**
The script makes the following assumptions:
1. The directory containing the input burst is organized by date. This should be a pretty common assumption for a stack processor

2. Metadata are in `json` format and it contains info on the `granule_id` (i.e. filename), `date`, `polygon`, burst_id, and `epsg`. These should be all pretty reasonable assumptions

3. Untested: the algorithm should equivalently work also for range/Doppler co-registered stacks of bursts, assuming that the same metadata (e.g. polygon of valid pixels) are provided.

**Testing**
The PR has been tested on a small stack of S1-A/B data that can be found on the aurora server at: `/mnt/aurora-r0/vbrancat/data/S1/gburst_stitching/ `. Below a pic of an interferogram formed by randomly selecting a reference and a secondary stitched SLC from the processed stack.

**TO DO**
1. Check if `gdal.Warp` allows to allocate the number of threads to use and expose it in the command line option
2. Expose in command line options X/Y resolution of stitched bursts and EPSG for reprojection
5. Check that the algorithm applies to stitch multiband rasters (e.g. geocoded bursts with main, low, and high bands

Stitching


src/compass/utils/stitching/stitch_burst.py
```diff
@@ -0,0 +1,404 @@
+import argparse
+import glob
+import json
+import os
+import time
+
+import isce3
+import journal
+import pandas as pd
```
LiangJYu, line 9: @vbrancat do you anticipate pandas being used outside burst stitching? I'm split between using pandas to make it super easy to select/filter bursts/dates and keeping dependencies light by writing some not-too-complex logic.

vbrancat: @LiangJYu I would like to use `pandas` also in the geocoded and radar burst stack processors. I find it very handy in handling the bursts to process. Using `pandas` should be simple and fast enough to do not require writing any additional logic. Not sure if @hfattahi has a strong opinion on this matter.

hfattahi: I agree with Virginia. We can hide the pandas dependency for SDS deliveries as they won't use the stack processor or stitcher.

src/compass/utils/stitching/stitch_burst.py
```diff
@@ -0,0 +1,404 @@
...
+
+    # Collect info for stitching in all dirs in 'indir'
+    # and return a panda dataframe with info
+    data_dict = get_stitching_dict(indir)
+
+    # If stitching some bursts, prune dataframe to
+    # contains only the burst IDs to stitch
+    if burst_list is not None:
+        data_dict = prune_dataframe(data_dict,
+                                    'burst_id', burst_list)
```
LiangJYu, line 95: This can be removed if suggestion to filter burst IDs in `get_stitching_dict`/`get_stitching_dataframe` accepted.

src/compass/utils/stitching/stitch_burst.py
```diff
@@ -0,0 +1,404 @@
...
+
+    # Identify common burst IDs among different dates
+    ids2stitch = get_common_burst_ids(data_dict)
+
+    # Prune dataframe to contain only the IDs to stitch
+    data_dict = prune_dataframe(data_dict,
+                                'burst_id', ids2stitch)
```
LiangJYu, line 102: If `prune_dataframe` above is removed, then `get_common_burst_ids` and `prune_dataframe` can be merged into `prune_uncommon_burst_ids` function? Unless you see `prune_dataframe` being being something to be imported and used elsewhere...

Something like:
```python
def prune_uncommon_burst_ids(data):
    '''
    Prune dataframe based on column ID and list of value
    Parameters:
    ----------
    data: pandas.DataFrame
       dataframe that needs to be pruned

    Returns:
    -------
    data: pandas.DataFrame
       Pruned dataframe with rows in 'id_list'
    '''
    unique_dates = list(set(data['date']))

    # Initialize list of unique burst IDs
    common_id = data.burst_id[data.date == unique_dates[0]]

    for date in unique_dates:
        ids = data.burst_id[data.date == date]
        common_id = sorted(list(set(ids.tolist()) & set(common_id)))

    # remove burst IDs not common to all days
    pattern = '|'.join(common_id)
    dataf = data.loc[data['burst_id'].str.contains(pattern,
                                               case=False)]
    return dataf
```

src/compass/utils/stitching/stitch_burst.py
```diff
@@ -0,0 +1,404 @@
...
+        # List metadata files in the directory
+        meta_list = sorted(glob.glob(f'{indir}/{dir}/*json'))
+        for path in meta_list:
+            # Read metadata file
+            metadata = read_metadata(path)
+            # Read info and store in dictionary
+            cfg['burst_id'].append(get_metadata(metadata, 'burst_id'))
+            datestr = get_metadata(metadata, 'sensing_start')
+            date = datetime.fromisoformat(datestr).strftime("%Y%m%d")
+            filename = f"{get_metadata(metadata, 'burst_id')}_{date}_VV.slc"
```
LiangJYu, line 297: Add a polarization option?

src/compass/utils/stitching/stitch_burst.py
```diff
@@ -0,0 +1,404 @@
...
+import time
+
+import isce3
+import journal
+import pandas as pd
+import shapely.wkt
+from datetime import datetime
+from compass.utils import helpers
+from osgeo import gdal, ogr
+from shapely.geometry import Polygon
```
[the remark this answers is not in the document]
vbrancat: Just heads-up, using `isort` (package to sort imports according to `pep8` convention) gives me a slightly different result.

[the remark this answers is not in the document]
LiangJYu: The `isort` output appears to group `from x import y` incorrectly. e.g. `from datetime import datetime` should be with the standard library group.

src/compass/utils/stitching/stitch_burst.py
```diff
@@ -0,0 +1,404 @@
...
+            cfg['burst_id'].append(get_metadata(metadata, 'burst_id'))
+            datestr = get_metadata(metadata, 'sensing_start')
+            date = datetime.fromisoformat(datestr).strftime("%Y%m%d")
+            filename = f"{get_metadata(metadata, 'burst_id')}_{date}_VV.slc"
+            cfg['granule_id'].append(f'{indir}/{dir}/{filename}')
+            poly = get_metadata(metadata, 'border')
+            cfg['polygon'].append(shapely.wkt.loads(poly))
+            cfg['date'].append(date)
+            geogrid = get_metadata(metadata, 'geogrid')
+            cfg['epsg'].append(geogrid['epsg'])
```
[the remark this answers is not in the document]
LiangJYu: This could probably be faster by filtering by file name ([file name includes `burst_id`](https://github.com/opera-adt/COMPASS/blob/6d5f80247a0baf1b146673471543cb34dc7ac6e2/src/compass/geo_cslc.py#L125)) and skipping the file loading/reading [here](https://github.com/opera-adt/COMPASS/blob/6d5f80247a0baf1b146673471543cb34dc7ac6e2/src/compass/utils/stitching/stitch_burst.py#L292). e.g.
```python
        for path in meta_list:
            # Check if burst_id wanted
            if path.split('_')[0] not in wanted_burst_ids:
                continue
```

src/compass/utils/stitching/stitch_burst.py
```diff
@@ -0,0 +1,404 @@
...
+            poly = Polygon(list(zip(tgt_x, tgt_y)))
+        poly_int = poly.intersection(poly_int)
+
+    # To be conservative, apply some margin to the polygon (TO DO:
+    # check eps)
+    poly_int = poly_int.buffer(-margin)
+    return poly_int, epsg_int
+
+
+def get_stitching_dict(indir):
```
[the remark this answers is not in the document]
LiangJYu: I made my suggestion because [this block of code](https://github.com/opera-adt/COMPASS/blob/6d5f80247a0baf1b146673471543cb34dc7ac6e2/src/compass/utils/stitching/stitch_burst.py#L89-L95):
```python
    data_dict = get_stitching_dict(indir)

    # If stitching some bursts, prune dataframe to
    # contains only the burst IDs to stitch
    if burst_list is not None:
        data_dict = prune_dataframe(data_dict,
                                    'burst_id', burst_list)
```
By [filtering out unwanted bursts in `get_stitching_dict`](https://github.com/opera-adt/COMPASS/pull/14#discussion_r915167917), the call to `prune_dataframe` is no longer needed and other calls to `prune_dataframe` will work on **smaller** dataframes. I wager a smaller `dataframe` will be faster to filter. If `burst_list is None`, downstream behavior is not impacted since everything in `indir`will be in `data_dict`.

My perspective of filtering is of what's present in the code of this PR. What others besides the 2 here that you have in mind?

src/compass/utils/stitching/stitch_burst.py
LiangJYu: Thank you for pointing out commonality with #18; I completely missed this. In light of this, I think `prune_dataframe` and `get_common_burst_ids` could be in their own file in src/compass/utils/dataframe_tools.py as a shared import between stack processing and stitching.

vbrancat:
@scottstanie Maybe we can close this PR? The stitching code has been incorporated in `dolphin` and it is much easier :)

scottstanie:
Sure! There's a chance that we later come across some case where stitching geocoded SLCs leads to an easier time... but for the current effort of the displacement workflow you're right that we've got that now in dolphin 👍

vbrancat:
Close as duplicated in the displacement workflow

PR #15, opened 2022-04-18
Logging formatting (pyre 1.11.0)

vbrancat:
This PR creates and format each entry of the log file at a Python level as:

`<time-tag>, <log level>, <workflow>, <module>, <error code>, <error location>, <logged message>`

where:

- `<time-tag>` is the time the message was logged, should be in ISO format (YYYY-MM-DDTHH:MMSS.mmmmmmZ)
- `<log level>` is the level of the log message, typically “Debug”, “Info”, “Warning”, or “Critical”
- `<workflow>` is typically the name of the SAS or PGE, so in this case “CSLC” or similar
- `<module>` is the name of the Python module where the logging took place
- `<error code>` is an integer code associated to the logged message. You pretty much have free range to define these, but they should be documented somewhere.
- `<error location>` is the file path of the Python module that did the logging, along with the line number concatenated with a colon (ex `/home/conda/opera/COMPASS/rdr2geo.py:645`)
- `<logged message>` is the actual logged message content

Note that this solution is temporary. With `pyre 1.11.2` and the use of the newly implemented NISAR device inside pyre should automatically format the logging message as above with the exception of assigning an error code. 

**TO DO**:

1. Assign error codes for Warnings, Critical etc for ISCE3 related errors (deserves group discussion)

### DIFF
--- a/.idea/COMPASS.iml
+++ b/.idea/COMPASS.iml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$">
+      <excludeFolder url="file://$MODULE_DIR$/venv" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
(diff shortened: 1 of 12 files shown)


vbrancat:
Closing since we decided to move forward without `pyre` logging

PR #18, opened 2022-05-10
WIP: Geocoded burst stack processor

vbrancat:
This PR adds a script for the geocoded burst stack processor.

The script works on a static stack of S1-A/B SAFE files and operates at a burst level.
Given a reference acquisition date in the stack, the script identifies the common bursts among the static stack of SAFE files and generate a runconfig and a bash file to use to run the geocoded burst workflow. 

Stack of burst with the same burst IDs are geocoded on the same grid which their stitching easier.
If the orbit directory is not provided, the script will automatically download the orbit ephemerides files for each SAFE file of the stack.

TO DO:
1. Have the script be capable of reading options from a configuration file


hfattahi:
@vbrancat I don't think if you have pushed any code. Would you please check and push the code?

hfattahi:
@vbrancat now that the repo reorganization is merged, would you please rebase so that we can start evaluating this PR?

vbrancat:
@hfattahi let me try to test the branch. I will push all the updates when ready :s

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,510 @@
...
+                                                      work_dir, burst_map, flatten,
+                                                      is_split_spectrum,
+                                                      low_band, high_band, pol,
+                                                      x_spac, y_spac)
+                    date_str = burst.sensing_start.strftime("%Y%m%d")
+                    runfile_name = f'{run_dir}/run_{date_str}_{burst.burst_id}.sh'
+                    with open(runfile_name, 'w') as rsh:
+                        path = os.path.dirname(os.path.realpath(__file__))
+                        rsh.write(
+                            f'python {path}/geo_cslc.py {runconfig_path}\n')
```
hfattahi, line 497:
```suggestion
                            f'python {path}/s1_cslc.py {runconfig_path}\n')
```

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,510 @@
...
+    for safe in zip_file:
+        i_subswath = [1, 2, 3]
+        orbit_path = get_orbit_file_from_list(safe,
+                                              glob.glob(f'{orbit_dir}/S1*'))
+
+        for subswath in i_subswath:
+            bursts = load_bursts(safe, orbit_path, subswath)
+            for burst in bursts:
+                if burst.burst_id in list(set(burst_map['burst_id'])):
+                    runconfig_path = create_runconfig(burst, safe, orbit_path,
```
hfattahi, line 486: Currently runconfig_path is relative path. It would be great to make it absolute path

hfattahi:
Very exciting PR. Before dig into the code, I'm trying to be a user and giving it a try. Provided two minor comments above. One nice functionality would be the start and end date. Let's assume there is a folder with all SLCs, but we want to process only certain dates. It would be nice to have that flexibility. Other relevant functionality would be allowing to process a list of specific dates. Or exclude a list of specific dates.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,510 @@
...
+                                                      dem_file,
+                                                      work_dir, burst_map, flatten,
+                                                      is_split_spectrum,
+                                                      low_band, high_band, pol,
+                                                      x_spac, y_spac)
+                    date_str = burst.sensing_start.strftime("%Y%m%d")
+                    runfile_name = f'{run_dir}/run_{date_str}_{burst.burst_id}.sh'
+                    with open(runfile_name, 'w') as rsh:
+                        path = os.path.dirname(os.path.realpath(__file__))
+                        rsh.write(
```
hfattahi, line 496: This currently creates many run files which each of them has only one command to run. An alternative would be to have one run file with many commands at each line. The latter allows to run multiple commands in parallel for example using [this](https://github.com/isce-framework/isce2/blob/main/contrib/stack/topsStack/run.py). I wonder if it is preferred to have one run file with one command per line or multiple run files with one command per file?
Maybe @seongsujeong has a preference?

vbrancat: @seongsujeong before I make any modification, what is your opinion with respect to the comment above?

vbrancat:
@hfattahi latest commit should address the following:

1. Inspect the stack and identifying the common bursts along the stack. If no `burst_id` is specified for processing, only the common bursts are used
2. Specify a `start_date` and an `end_date`. If specified, all the burst IDs with dates in this interval will be processed
3. Ability to exclude dates from the stack 

I should do some code clean-up (move orbits download to the reader), improve panda dataframe filtering (it is a bit slow)

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,566 @@
...
+    burst_map = generate_burst_map(zip_file, orbit_dir, x_spac, y_spac, epsg)
+
+    # Identify burst IDs common across the stack and remove from the dataframe
+    # burst IDs that are not in common
+    common_ids = get_common_burst_ids(burst_map)
+    burst_map = prune_dataframe(burst_map, 'burst_id', common_ids)
+
+    # If user selects burst IDs to process, prune unnecessary bursts
+    if burst_id is not None:
+        burst_map = prune_dataframe(burst_map, 'burst_id', burst_id)
```
LiangJYu, line 514: Can avoid this `prune_dataframe` call if  `generate_burst_map` takes in `burst_id` and uses it to filter bursts. Like [this](https://github.com/opera-adt/COMPASS/pull/14#discussion_r915167917).

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,566 @@
...
+
+    # Select only dates between start and end
+    if start_date is not None:
+        burst_map = burst_map[burst_map['date'] >= start_date]
+    if end_date is not None:
+        burst_map = burst_map[burst_map['date'] <= end_date]
+
+    # Exclude some dates if the user requires it
+    if exclude_dates is not None:
+        burst_map = prune_dataframe(burst_map, 'date', exclude_dates)
```
LiangJYu, line 524: Like `burst_id`, these bits of pruning could be filtered in `generate_burst_map` as well.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,566 @@
...
+
+    # Ready to geocode bursts
+    for safe in zip_file:
+        i_subswath = [1, 2, 3]
+        orbit_path = get_orbit_file_from_list(safe,
+                                              glob.glob(f'{orbit_dir}/S1*'))
+        for subswath in i_subswath:
+            bursts = load_bursts(safe, orbit_path, subswath)
```
LiangJYu, line 532: If [`Sentinel1BurstSlc`](https://github.com/opera-adt/s1-reader/blob/e642bd645aea820e03a3497df9de1c647cb71f0a/src/s1reader/s1_burst_slc.py#L127) objects are stored in `burst_map` then they don't need to reloaded here.

If my suggestions to filter in `generate_burst_map` are accepted, then that should keep the `burst_map` object not too big?

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,566 @@
...
+                if (burst.burst_id in list(set(burst_map['burst_id']))) and \
+                        (date in list(set(burst_map['date']))):
+                    runconfig_path = create_runconfig(burst, safe, orbit_path,
+                                                      dem_file,
+                                                      work_dir, burst_map,
+                                                      flatten,
+                                                      is_split_spectrum,
+                                                      low_band, high_band, pol,
+                                                      x_spac, y_spac)
+                    date_str = str(date)
```
scottstanie, line 544: I think this line should be moved to right after line 534. Right now if the "else" triggers, the logging entry on line 552 uses `date_str`, but it isn't defined.

src/compass/s1_geo_stack.py
scottstanie: The start and end date pruning is now occurring before the generation, so that the number of zip files to load is minimized

src/compass/s1_geo_stack.py
scottstanie: This suggestions is implemented in https://github.com/opera-adt/COMPASS/pull/53 , and it means that the script run time is cut in half (since 95% or more of the script time is spent loading, and every zip was loaded twice)

src/compass/s1_geo_stack.py
scottstanie: I've been running them in parallel like 
```
ls run_files  | xargs -P 10 sh -c 'bash run_files/$0 > logfile-$0.log'
```
so for now I've kept them as separate files

scottstanie:
I believe I've addressed the comments here, so moving the final implementation to https://github.com/opera-adt/COMPASS/pull/53

PR #20, opened 2022-05-11
WIP rdr2geo + geo2rdr unit tests

LiangJYu:
This PR is based the [rdr2geo and geo2rdr unit tests](https://gist.github.com/LiangJYu/fb7c4c9be0ba21bd1dc5c82f66dbce80) @hfattahi created for the now deprecated Cython ISCE3. All test inputs are generated by the test. The objective for the this PR is have the 4 following tests:

1. rdr2geo with satellite orbit over equator and zero DEM with zero DEM
2. rdr2geo with satellite orbit over constant 45° lat with zero DEM
3. rdr2geo with satellite orbit over equator and zero DEM with zero DEM
4. rdr2geo with satellite orbit over constant 45° lat with zero DEM

TBD:

- Decide whether to use existing `SimpleNamespace` to mimic `RunConfig` shortcut or introduce test yamls and SAFE zip for comprehensiveness.

To do:

- debug so rdr2geo test passes
- implement both equator and 45° lat geo2rdr tests
- more documentation and comments

### DIFF
--- a/tests/prep.py
+++ b/tests/prep.py
@@ -1,0 +1,180 @@
+import datetime
+import pytest
+from types import SimpleNamespace
+
+import isce3
+import numpy as np
+from osgeo import gdal
+from s1reader.s1_burst_slc import Doppler, Sentinel1BurstSlc
+
+def zero_dem(params):
+    lon0 = params.lon0 - params.omega / 2
+    lat0 = params.lat0 - params.omega / 2
+    geo_trans = [lon0, params.omega, 0.0, lat0, 0.0, params.omega]
+
+    length = 1
+    width = params.n_sv + 1
+
+    # create DEM file
+    dem_raster = isce3.io.Raster(params.dem, width, length, 1,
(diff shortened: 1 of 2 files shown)


LiangJYu:
The rdr2geo test is currently failing because the DEM interpolator is returning the wrong height. The indices computed by DEM interpolator do not fall within the synthetic DEM so the [default value (DEM average) is returned](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/4e4fe4ad9e5957d012ab34d64e013abfc06c2bc5/cxx/isce3/geometry/DEMInterpolator.cpp#L605) for all points in the radar grid. I think a couple things need addressing:
- a DEM that fits the radar grid. **I'm not sure how to do this.**
- ability to change DEM interp method to nearest neighbor so the DEM can remain small. This I can fix.

Since the Cython rdr2geo test tests point by point, it avoids this problem by changing the DEM interpolator default value to expected height per point.

Issue #16, opened 2022-04-21
repo structure


yunjunz:
Seeing the growing COMPASS repo, I am thinking of the following changes:

1. rename configuration files from `cslc_s1.yaml` to `s1_cslc.yaml`, to be consistent with the script name `s1_cslc.py`
2. move `schemas` and `defaults` from within `utils` to the same level as `utils` for shoter path, since users will check them frequently
3. create a `bin` directory for executables, like `s1_cslc.py`, as they are really executable scripts. Update: we could add entry_points for executables, without moving them to a `bin` folder.

These lead to a structure as below:

```bash
/COMPASS
    /bin
        s1_cslc.py
        s1_gslc.py
    /src/compass
        s1_rdr2geo.py
        s1_geo2rdr.py
        s1_resample.py (renamed from s1_resample_burst.py)
        ...
        /utils
            helpers.py
            geo_grid.py (renamed from geogrid.py)
            radar_grid.py (renamed from reference_radar_grid.py)
            runconfig.py
            ...
        /defaults
            s1_cslc.yaml (renamed from cslc_s1.py)
            s1_gslc.yaml (renamed from geo_cslc_s1.yaml)
        /schemas
            s1_cslc.yaml
            s1_gslc.yaml
    /tests
        ...
```

Let me know what do you think please @LiangJYu @vbrancat @hfattahi .

hfattahi:
I like the idea. For OPERA we chose the term CSLC to represent Coregistered SLC in general regardless if they are in radar or geo-coordinates. If we want to avoid confusion in that regard, then I suggest to use:

```
s1_cslc_radar.py
s1_cslc_geo.py
```

That being said, the name that I really like for GSLC is ```geocode_slc.py```

yunjunz:
> That being said, the name that I really like for GSLC is `geocode_slc.py`

I assume you meant `s1_geocode_slc.py`? I like it a lot too. 

If the geocode SLC executable script can be as short as the current `s1_cslc.py`, we can add merge it into `s1_cslc.py` and add an option in the yaml file to choose between geo or radar grid. That would be even simpler. And we could keep one yaml file also, instead of two.

LiangJYu:
I like the changes. A few questions/comments:

How does `bin` get installed? I have no idea and an explanation would be very useful. I've been installing COMPASS into my conda env and running with `python -m` and would like to learn more.

Regarding the common executable, were you thinking the following?
```
/COMPASS
    /bin
[...]

PR #21, opened 2022-05-12
re-organize repo structure

yunjunz:
This PR implements the code structure as discussed in #16 via renaming scripts and moving the locations of `schemas` and `defaults` folders. It also includes the following changes:

+ refactor `s1_cslc.py` to support both geo and radar CSLC workflow. Two questions:
   ~~1. The workflow type is identified from the input yaml file via keys in `runconfig/groups/processing`. Let me know if there is an easier or more robust way please.~~
   2. It will be helpful to have command-line examples print out while users type `s1_cslc.py -h`. This is done in mintpy via `argparse.ArgumentParser(epilog)`. But I don't know how to implement this here since all the current scripts under `src/compass` are sharing the same `yaml_argparse.py`. One solution is to use `argparse.add_argument_group()` for re-using the same cmd options, but still keeps the flexibility/variability needed for each script, mintpy did this [here](https://github.com/insarlab/MintPy/blob/7014b0b328da5819374dae01e8defd96b6d42646/mintpy/utils/arg_group.py#L51) and [here](https://github.com/insarlab/MintPy/blob/7014b0b328da5819374dae01e8defd96b6d42646/mintpy/view.py#L124). Let me know what do you think please.

+ add an `entry_points` to `s1_cslc:main()` to make the script available on command line.

+ for the radar CSLC workflow, use YYYYMMDD as date_str w/o hyphen, for the folder name.

+ detailed renaming changes are as below:

```bash
/COMPASS
    /src/compass
        s1_cslc.py
        s1_rdr2geo.py
        s1_geo2rdr.py
        s1_geocode_slc.py (renamed from geo_cslc.py)
        s1_resample.py (renamed from s1_resample_burst.py)
        ...
        /utils
            helpers.py
            geo_grid.py (renamed from geogrid.py)
            geo_metadata.py (renamed from geo_cslc_metadata.py)
            radar_grid.py (renamed from reference_radar_grid.py)
            runconfig.py
            del unwrap_namespace.py (duplicate to wrap_namespace.py)
            ...
        /defaults
            s1_cslc_radar.yaml (renamed from cslc_s1.py)
            s1_cslc_geo.yaml (renamed from geo_cslc_s1.yaml)
        /schemas
            s1_cslc_radar.yaml
            s1_cslc_geo.yaml
```


src/compass/s1_cslc.py
```diff
@@ -1,25 +1,50 @@
-from compass import s1_rdr2geo, s1_geo2rdr, s1_resample_burst
+from ruamel.yaml import YAML
```
LiangJYu, line 1: I'm unsure about a top level function that consolidates radar and geo workflows as-is. I see the value in its simplicity. The yaml parsing is redundant and without error checking. I haven't been able to find a good workaround when the high level workflow wrapper is takes only **one** argument to cfg yaml. Taking another argument for rdr or geo mode makes yaml parsing in the file and all the text/code below irrelevant. 🤣 

I tried the following wrapper around both RunConfigs so the yaml is only read once but I am unable to come up with a way to capture the yamale validation failures because the stack traces leave out validation errors. Even if validation errors from Yamale can be captured when validation fails, the output would messy and ambiguous.

```python
import sys

import yamale

from compass.utils.geo_runconfig import GeoRunConfig
from compass.utils.runconfig import RunConfig


def cfg_load(f_yaml):
    cfg = {}
    mode = None
    e_rdr = None
    e_geo = None
    try:
        # try loading radar cfg
        cfg = RunConfig.load_from_yaml(f_yaml, 's1_cslc_rdr')
        mode = 'rdr'
    except Exception as e_rdr:
        pass

    try:
        # try loading geo cfg
        cfg = GeoRunConfig.load_from_yaml(f_yaml, 's1_cslc_geo')
        mode = 'geo'
    except Exception as e_geo:
        pass

    if mode is None:
        # print both stack traces
        raise ValueError

    return cfg, mode

if __name__ == "__main__":
    f_yaml = sys.argv[1]
    cfg = cfg_load(f_yaml)
    if cfg:
        print('winner')
    else:
        print('loser')
```

setup.cfg
```diff
@@ -24,3 +24,8 @@ install_requires =
 
 [options.packages.find]
 where = src
+
+[options.entry_points]
+console_scripts =
+    s1_cslc.py = compass.s1_cslc:main
```
LiangJYu, line 30: Nice! This is all it takes to do the following in terminal?
```bash
random_path$ c1_cslc some_yaml
```

LiangJYu (commented):
Nice work! I like what you've done. Just to couple issues that shouldn't be big deals.

setup.cfg
yunjunz: yep, and without `random_path$`.

src/compass/s1_geo2rdr.py
```diff
@@ -87,17 +87,16 @@ def run(cfg: dict):
...
 
 if __name__ == "__main__":
     """Run geo2rdr from command line"""
-    geo2rdr_parser = YamlArgparse()
+    parser = YamlArgparse()
 
     # Get a runconfig dict from command line arguments
-    geo2rdr_runconfig = RunConfig.load_from_yaml(
-        geo2rdr_parser.args.run_config_path, 'geo2rdr')
+    cfg = RunConfig.load_from_yaml(parser.args.run_config_path, 's1_geo2rdr')
```
yunjunz, line 99: I don't quite understand this part. Should this line be as below for `s1_geo2rdr`, `s1_rdr2geo` and `s1_resample`?

```
cfg = RunConfig.load_from_yaml(parser.args.run_config_path, 's1_cslc_radar')
```

src/compass/s1_cslc.py
yunjunz: I like the idea of having a cmd option, e.g. `-w/--workflow geo/radar`. But this does not seem doable together with `YamlArgparse()`. What I really like would be an independent, like the [`create_parser()` in `s1_geo_stack()`](https://github.com/vbrancat/COMPASS/blob/63782c0dc08dc961c33a0f62ab88c47927101729/src/compass/s1_geo_stack.py#L34). It's more flexible, and more importantly, extendable in the long term if we would like to add more options in the future. Shall we go for that?

src/compass/s1_geo2rdr.py
LiangJYu: > I don't quite understand this part. @vbrancat @LiangJYu Should this line be as below for `s1_geo2rdr`, `s1_rdr2geo` and `s1_resample`?
> 
> ```
> cfg = RunConfig.load_from_yaml(parser.args.run_config_path, 's1_cslc_radar')
> ```

[`load_from_yaml`](https://github.com/yunjunz/COMPASS/blob/70633bcf290e69339d789cbb910092efeb9fb777/src/compass/utils/runconfig.py#L310) passes the 2nd parameter to [`load_validate_yaml`](https://github.com/yunjunz/COMPASS/blob/70633bcf290e69339d789cbb910092efeb9fb777/src/compass/utils/runconfig.py#L35). In `load_validate_yaml`, the 2nd parameter/`workflow_name` is used to determine `schema_name`.

src/compass/s1_cslc.py
LiangJYu: How do you like the following? https://github.com/yunjunz/COMPASS/pull/1

src/compass/s1_geo2rdr.py
yunjunz: Thank you @LiangJYu, I will change the arg input for s1_geo2rdr/rdr2geo/resample then.

src/compass/s1_cslc.py
yunjunz: Thank you @LiangJYu, I like it. I moved the call of YamlArgparse into the main() as the `entry_points` prefers an arg-less main().

src/compass/s1_cslc.py
```diff
@@ -1,25 +1,35 @@
...
-def run(cfg):
-    # If boolean flag "is_reference" is true
-    # Run rdr2geo and archive reference burst
-    if cfg.is_reference:
-        s1_rdr2geo.run(cfg)
-    else:
-        s1_geo2rdr.run(cfg)
-        s1_resample_burst.run(cfg)
+def main():
+    parser = YamlArgparse(add_grid_type=True)
```
LiangJYu, line 8: I'm prefer to the putting `YamlArgparse` in `__main__` as this allows `s1_cslc.main` to be called from another function. Like in the [here](https://github.com/yunjunz/COMPASS/pull/1/commits/6932068270c549b9d0d0f32493508823458004bb).

yunjunz: The YamlArgparse() is needed inside the main() to parse the command line inputs too, as the `entry_points` in `setup.cfg` points to the `main()` while running s1_cslc.py as executables. And the entry_points could not be `__main__`. 

I do like the call to main() from other functions as well. A simple solution to achieve both is: create a `run(cfg, grid_type)` and move most of the code inside main() there. Then we could call `s1_cslc.run()` in a similar way to other scripts in python. Do you like this version below?

```
def run(run_config_path: str, grid_type: str):
    if grid_type == 'radar':
        # CSLC workflow in radar coordinates
        # get a runconfig dict from command line args
        cfg = RunConfig.load_from_yaml(run_config_path, 's1_cslc_radar')

        if cfg.is_reference:
            # reference burst - run rdr2geo and archive it
            s1_rdr2geo.run(cfg)

        else:
            # secondary burst - run geo2rdr + resample
            s1_geo2rdr.run(cfg)
            s1_resample.run(cfg)

    elif grid_type == 'geo':
        # CSLC workflow in geo-coordinates
        # get a runconfig dict from command line argumens
        cfg = GeoRunConfig.load_from_yaml(run_config_path, 's1_cslc_geo')

        # run geocode_slc
        s1_geocode_slc.run(cfg)

def main():
    parser = YamlArgparse(add_grid_type=True)

    run(parser.run_config_path, parser.args.grid_type)
```

LiangJYu: LGTM

LiangJYu (approved):
LGTM

hfattahi (approved):
LGTM! Thanks @yunjunz . I was able to run the geocoded workflow.

vbrancat (approved):
LGTM. I also ran the geocoded workflow without any issue. Thanks a lot for having done this re-organization :)

PR #22, opened 2022-05-13
Change geocoded burst Interface

vbrancat:
This PR includes changes to the interface of the geocoded burst workflow.
The interface now includes paths to:

- annotation file
- calibration file
- rfi file
- noise file
- burst data path
- light metadata path

### DIFF
--- a/src/compass/utils/defaults/geo_cslc_s1.yaml
+++ b/src/compass/utils/defaults/geo_cslc_s1.yaml
@@ -10,16 +10,8 @@ runconfig:
           safe_file_path:
           # Required. List of orbit (EOF) files (min=1)
           orbit_file_path:
-          # Required. Path to annotation file
-          annotation_file_path:
-          # Required. Path to RFI file
-          rfi_file_path:
-          # Required. Path to noise file
-          noise_file_path:
-          # Required. Path to calibration file
-          calibration_file_path:
           # Required. Path to the burst data
-          burst_file_path: 
+          burst_file_path:
           # Required. Path to light metadata file
           ligh_metadata_path:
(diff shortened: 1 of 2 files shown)


src/compass/utils/defaults/geo_cslc_s1.yaml
```diff
@@ -10,6 +10,18 @@ runconfig:
...
           # Required. List of orbit (EOF) files (min=1)
           orbit_file_path:
+          # Required. Path to annotation file
+          annotation_file_path:
+          # Required. Path to RFI file
+          rfi_file_path:
+          # Required. Path to noise file
+          noise_file_path:
+          # Required. Path to calibration file
+          calibration_file_path:
```
hfattahi, line 20: @vbrancat as we discussed offline, I think the easiest approach would be to not expose any of these four files for IF delivery. They all exist in the  safe_file_path for now.

hfattahi (approved):
Thanks @vbrancat. LGTM!

src/compass/utils/defaults/geo_cslc_s1.yaml
```diff
@@ -10,6 +10,10 @@ runconfig:
           safe_file_path:
           # Required. List of orbit (EOF) files (min=1)
           orbit_file_path:
+          # Required. Path to the burst data
+          burst_file_path:
```
LiangJYu, line 14: `burst_file_path` is the ASF file?

hfattahi: It will be any gdal readable file that contains the data for one burst.

yunjunz (approved):
LGTM.

PR #29, opened 2022-05-24
download orbit

hfattahi:
Borrowing the fetchorbit script from our isce2 stack processor and slightly modifying so that a directory with all safe files can be given as input to loop over and download the orbits for.

### DIFF
--- a/src/compass/download_orbits.py
+++ b/src/compass/download_orbits.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python3
+
+import numpy as np
+import requests
+import re
+import os
+import glob
+import argparse
+import datetime
+from html.parser import HTMLParser
+
+server = 'https://scihub.copernicus.eu/gnss/'
+
+orbitMap = [('precise', 'AUX_POEORB'),
+            ('restituted', 'AUX_RESORB')]
+
+datefmt = "%Y%m%dT%H%M%S"
+queryfmt = "%Y-%m-%d"
+queryfmt2 = "%Y/%m/%d/"
+
+#Generic credentials to query and download orbit files
+credentials = ('gnssguest', 'gnssguest')
+
+
+def cmdLineParse():
+    '''
+    Command line parser.
+    '''
+
+    parser = argparse.ArgumentParser(description='Fetch Sentinel-1 orbits')
+    parser.add_argument('-i', '--input-dir', dest='input_dir', type=str, required=True,
+                        help='Path to the directory with all Sentinel-1 SAFE files')
+    parser.add_argument('-o', '--output', dest='outdir', type=str, required=True,
+                        help='Path to output orbit directory')
+
+    return parser.parse_args()
+
+
+def FileToTimeStamp(safename):
+    '''
+    Return timestamp from SAFE name.
+    '''
+    safename = os.path.basename(safename)
+    fields = safename.split('_')
+    sstamp = []  # sstamp for getting SAFE file start time, not needed for orbit file timestamps
+
+    try:
+        tstamp = datetime.datetime.strptime(fields[-4], datefmt)
+        sstamp = datetime.datetime.strptime(fields[-5], datefmt)
+    except:
+        p = re.compile(r'(?<=_)\d{8}')
+        dt2 = p.search(safename).group()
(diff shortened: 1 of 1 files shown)


src/compass/download_orbits.py
```diff
@@ -0,0 +1,178 @@
...
+
+def fetch_orbit(safe_file, orbit_dir):
+    fileTS, satName, fileTSStart = FileToTimeStamp(safe_file)
+    print('Reference time: ', fileTS)
+    print('Satellite name: ', satName)
+    match = None
+    session = requests.Session()
+
+    for spec in orbitMap:
+        oType = spec[0]
```
vbrancat, line 131:
```suggestion
        o_type = spec[0]
```

src/compass/download_orbits.py
```diff
@@ -0,0 +1,178 @@
...
+        success = False
+        match = None
+  
+        try:
+            r = session.get(url, verify=True, auth=credentials)
+            r.raise_for_status()
+            parser = MyHTMLParser(url)
+            parser.feed(r.text)
+            for resulturl, result in parser.fileList:
+                tbef, taft, mission = fileToRange(os.path.basename(result))
```
vbrancat, line 146:
```suggestion
                tbef, taft, mission = file_to_range(os.path.basename(result))
```

src/compass/download_orbits.py
```diff
@@ -0,0 +1,178 @@
...
+#Generic credentials to query and download orbit files
+credentials = ('gnssguest', 'gnssguest')
+
+
+def cmdLineParse():
+    '''
+    Command line parser.
+    '''
+
+    parser = argparse.ArgumentParser(description='Fetch Sentinel-1 orbits')
```
vbrancat, line 30: May be clarify the logic with which the orbits are downloaded, i.e. look for POE first if POE is not found download the ROE?

vbrancat (commented):
@hfattahi thanks a lot for these PR. My major comment is to try to uniform the coding style to pep8 when possible. Also while testing the script, I ran into an issue (solvable, I think by modifying the output path to which the orbits are allocated). Also, I think the script should go into the `utils` folder as it is a functionality more then a workflow. @LiangJYu and @yunjunz would for sure have an opinion on that.

yunjunz:
> @hfattahi thanks a lot for these PR. My major comment is to try to uniform the coding style to pep8 when possible. Also while testing the script, I ran into an issue (solvable, I think by modifying the output path to which the orbits are allocated). Also, I think the script should go into the `utils` folder as it is a functionality more then a workflow. @LiangJYu and @yunjunz would for sure have an opinion on that.

Since this script could be potentially added to a command-line executable, I like it under `src/compass`.

For the naming, shall we make it consistent with other scripts, e.g. `s1_dload_orbit.py`?

yunjunz:
For the convenience and flexibility on the user side, it will be great if we could support:

1. the input SLC path as safe file, in addition to the directory of it.
2. `--start` and `--end` date, without input safe file/dir, as what we have in `topsStack/dloadOrbit.py`.

LiangJYu:
I think this code is a better fit in [s1-reader](https://github.com/opera-adt/s1-reader) given the substantial amount of duplicate functionality in [s1_orbit.py](https://github.com/opera-adt/s1-reader/blob/main/src/s1reader/s1_orbit.py). @hfattahi if you agree, I will resubmit PR in s1-reader.

hfattahi:
> I think this code is a better fit in [s1-reader](https://github.com/opera-adt/s1-reader) given the substantial amount of duplicate functionality in [s1_orbit.py](https://github.com/opera-adt/s1-reader/blob/main/src/s1reader/s1_orbit.py). @hfattahi if you agree, I will resubmit PR in s1-reader.

That is fine. Also I came across this script as well https://github.com/scottstanie/sentineleof . It may does all we need

vbrancat:
@hfattahi we have moved the orbit download functionality to a new PR in the `s1-reader`. If you don't mind, I would be inclined to close this PR and move the conversation to the `s1-reader` repo. Keep the branch just in case. What do you think?

vbrancat:
Closing since functionality has been included in the s1-reader

PR #32, opened 2022-06-07
Support relative paths for SAFE files (needed for IF)

vbrancat:
This PR substitutes `os.path.isfile` with `os.path.exists` when checking for the validity of the input SAFE file. It seems that `os.path.isfile` does not support relative paths but would require to allocate full file path in the runconfig.


LiangJYu (approved):
Good find. Thanks for the fix!

seongsujeong (approved):
Good catch and fix.  👍👍👍

PR #33, opened 2022-06-07
Script to validate CSLC product & metadata (IF delivery)

vbrancat:
This PR introduces a script to validate generated CSLC products against a golden dataset. 
The script can be used by the PGE team to check their generated CSLC product with that provided by ADT.

The script performs checks on the:
- products: number of bands, geo transformation arrays, amplitude and phase difference
- metadata: number of metadata keys and their corresponding values


src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,144 @@
...
+        return
+
+    if not os.path.exists(file_2):
+        print(f'ERROR CSLC metadata not found: {file_2}')
+        return
+
+    # Load metadata
+    with open(file_1, 'r') as f:
+        metadata_1 = json.load(f)
+
```
LiangJYu, line 103:
```
        metadata_keys_ref = set(metadata_ref)
```
Cache instead of redoing. Also do the same for `_sec` below.

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,144 @@
...
+    with open(file_2, 'r') as f:
+        metadata_2 = json.load(f)
+
+    print('Compare number of metadata keys')
+    if not len(metadata_1.keys()) == len(metadata_2.keys()):
```
LiangJYu, line 108:
```suggestion
    if not len(metadata_keys_ref) == len(metadata_keys_sec):
```

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,144 @@
...
+        print('ERROR different number of metadata keys')
+        return
+
+    # Intersect metadata keys
+    set_1_m_2 = set(metadata_1.keys()) - set(metadata_2.keys())
+    if len(set_1_m_2) > 0:
```
LiangJYu, line 114:
```suggestion
    if set_ref_minus_sec:
```
Python sets, IIRC built-in containers in general, return `False` empty. Also rename for clarity.

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,144 @@
...
+    set_2_m_1 = set(metadata_2.keys()) - set(metadata_1.keys())
+    if len(set_2_m_1) > 0:
+        print(f'Secondary CSLC metadata has extra entries with keys:'
+              f'{", ".join(set_2_m_1)}.')
+
+    # Check remaining metadatakeys
+    for k1, v1 in metadata_1.items():
+        if k1 not in metadata_2.keys():
+            print(f'ERROR the metadata key {key1} in not present'
+                  f'in the secondary CSLC metadata')
```
LiangJYu, line 127:
```suggestion
```
Redundant - the check of set difference `set_ref_minus_sec` does the same above.

LiangJYu (commented):
Looks mostly good. Few comments are for clarity.

src/compass/utils/validate_cslc.py
vbrancat: Not sure this comment is doable. I get the following error with your suggestion:

```
  File "/u/aurora0/vbrancat/codes/validate_cslc.py", line 141, in <module>
    compare_cslc_metadata(cmd.ref_metadata, cmd.sec_metadata)
  File "/u/aurora0/vbrancat/codes/validate_cslc.py", line 116, in compare_cslc_metadata
    if set_ref_minus_sec > 0:
TypeError: '>' not supported between instances of 'set' and 'int'

```

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,142 @@
...
+    slc_ref = dataset_ref.GetRasterBand(1).ReadAsArray()
+    slc_sec = dataset_sec.GetRasterBand(1).ReadAsArray()
+
+    diff_amp = np.abs(slc_ref) - np.abs(slc_sec)
+    diff_pha = np.angle(slc_ref * np.conj(slc_sec))
+
+    print('Check max amplitude difference between CSLC products is < 1.0e-12')
+    assert np.nanmax(diff_amp) < 1.0e-12
+    print('Check max phase difference between CSLC products is < 1.0e-12')
+    assert np.nanmax(diff_pha) < 1.0e-12
```
hfattahi, line 82: My suggestion is to compare real and imag instead of phase and amp. 
Also relax the threshold to something around 1e-5 since we have single precision float data.

vbrancat: Done.  A question, do we want to compare that the max difference of the real (imaginary) part or rather than the mean value is < 1e-5?

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,143 @@
...
+    metadata_keys_ref = set(metadata_ref)
+
+    with open(file_sec, 'r') as f:
+        metadata_sec = json.load(f)
+    metadata_keys_sec = set(metadata_sec)
+
+    print('Compare number of metadata keys')
+    if not len(metadata_keys_ref) == len(metadata_keys_sec):
+        print('ERROR different number of metadata keys')
+        return
```
LiangJYu, line 113:
```suggestion
```
I don't think this is required given that 2 following checks do the same with greater detail.

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,143 @@
...
+    # Intersect metadata keys
+    set_ref_minus_sec = set(metadata_ref.keys()) - set(metadata_sec.keys())
+    if len(set_ref_minus_sec) > 0:
+        print(f'Reference CSLC metadata has extra entries with keys:'
+              f'{", ".join(set_ref_minus_sec)}.')
+        return
+    set_sec_minus_ref = set(metadata_sec.keys()) - set(metadata_ref.keys())
+    if len(set_sec_minus_ref) > 0:
+        print(f'Secondary CSLC metadata has extra entries with keys:'
+              f'{", ".join(set_sec_minus_ref)}.')
```
LiangJYu, line 124:
```suggestion
    # check Intersection of metadata keys
    set_ref_minus_sec = metadata_keys_ref - metadata_keys_sec
    set_sec_minus_ref = metadata_keys_sec - metadata_keys_ref
    
    # error string just in case
    err_str = "Metadata keys do not match.\n"
    if set_ref_minus_sec:
        err_str += f'Reference CSLC metadata extra entries: {set_ref_minus_sec }'
    if set_sec_minus_ref:
        err_str += f'Secondary CSLC metadata extra entries: {set_sec_minus_ref}'
        
    # check for keys differ
    assert(not set_ref_minus_sec or not set_sec_minus_ref, err_str)
```
Consolidate error check to show extras in both inputs. Also adds usage of cached keys.

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Check remaining metadatakeys
+    for k_ref, v_ref in metadata_ref.items():
+        if metadata_sec[k_ref] != v_ref:
+            print(f'ERROR the content of metadata key {k_ref} from'
+                  f'reference CSLC metadata has a value {v_ref} whereas the same'
+                  f'key in the secondary CSLC metadata has value {metadata_sec[k_ref]}')
```
LiangJYu, line 131: Is this meant to be an error string since no `assertion` is made?

The metadata between 2 bursts will definitely differ no?

LiangJYu (commented):
Few more changes. Let me know if my suggested set diff logic doesn't work and maybe we can debug over Slack.

src/compass/utils/validate_cslc.py
vbrancat: No the metadata between the two bursts will not differ. That is the point of the code. We are comparing a golden dataset (e.g. the one processed by ADT) with another another dataset assumed to be produced by the same runconfig (e.g. produced by the PGE team). So the metadata between the two should be consistent

vbrancat:
As a note, I changed the validation metric to use `numpy.allclose`. The difference we were seeing prior to IF delivery where due to CSLC datasets generated with slightly different version of ISCE3 and installed packages. When performing an apple-to-apple comparison, the generated datasets seem to show no difference.

hfattahi:
> As a note, I changed the validation metric to use `numpy.allclose`. The difference we were seeing prior to IF delivery where due to CSLC datasets generated with slightly different version of ISCE3 and installed packages. When performing an apple-to-apple comparison, the generated datasets seem to show no difference.

Thanks for the update!

vbrancat:
@LiangJYu I think I have address all your comments and this PR should be ready for a second round of reviews. @yunjunz or @seongsujeong does one of you have time to get this reviewed? I will gladly appreciate your inputs :)

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,136 @@
...
+def compare_cslc_products(file_ref, file_sec):
+    '''
+    Compare a reference and a newly generated
+    (i.e., secondary) CSLC product
+
+    Parameters
+    ----------
+    file1: str
+        File path to reference CSLC product (golden dataset)
+    file2: str
```
LiangJYu, line 37:
```suggestion
    file_ref: str
        File path to reference CSLC product (golden dataset)
    file_sec: str
```

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,136 @@
...
+    print('Check mean imaginary part difference between CSLC products is < 1.0e-5')
+    assert np.allclose(slc_ref.imag, slc_sec.imag,
+                       atol=0.0, rtol=1.0e-5, equal_nan=True)
+
+    return
+
+
+def compare_cslc_metadata(file_ref, file_sec):
+    '''
+    Compare reference and generated CSLC metadata
```
LiangJYu, line 88: Expand docstring with parameters.

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Intersect metadata keys
+    set_ref_minus_sec = metadata_ref.keys() - metadata_sec.keys()
+    set_sec_minus_ref = metadata_sec.keys() - metadata_ref.keys()
+
+    err_str = "Metadata keys do not match.\n"
+    if set_ref_minus_sec:
+        err_str += f'Reference CSLC metadata extra entries: {set_ref_minus_sec}'
+    if set_sec_minus_ref:
+        err_str += f'Secondary CSLC metadata extra entries: {set_sec_minus_ref}'
```
LiangJYu, line 115:
```suggestion
    err_str = "Metadata keys do not match."
    if set_ref_minus_sec:
        err_str += f'\nReference CSLC metadata extra entries: {set_ref_minus_sec}'
    if set_sec_minus_ref:
        err_str += f'\nSecondary CSLC metadata extra entries: {set_sec_minus_ref}'
```
Put everything on different line for readability.

LiangJYu (approved):
Few nits aside LGTM.

vbrancat:
@yunjunz and @seongsujeong do you have time to have a look at this PR? It should be good to go. I would gladly appreciate if one of you can have a look at it.

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,143 @@
...
+        return
+
+    # Extract some info from reference/secondary CSLC products
+    dataset_ref = gdal.Open(file_ref, gdal.GA_ReadOnly)
+    geotransform_ref = dataset_ref.GetGeoTransform()
+    nbands_ref = dataset_ref.RasterCount
+
+    dataset_sec = gdal.Open(file_sec, gdal.GA_ReadOnly)
+    geotransform_sec = dataset_sec.GetGeoTransform()
+    nbands_sec = dataset_sec.RasterCount
```
seongsujeong, line 57: Could be a nit-pick, but I would suggest taking care of the loaded gdal rasters (`dataset_ref` and `dataset_src`), or let python takes care of them, after done with them. For example,
```python
with gdal.Open(file_ref, gdal.GA_ReadOnly) as dataset_ref:
      geotransform_ref = dataset_ref.GetGeoTransform()
      nbands_ref = dataset_ref.RasterCount
      # Same for dataset_sec


```

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,143 @@
...
+        print(f'ERROR Number of bands in reference CSLC {nbands_ref} differs'
+              f'from number of bands {nbands_sec} in secondary CSLC')
+        return
+
+    print('Comparing geo transform arrays ...')
+    if not np.array_equal(geotransform_ref, geotransform_sec):
+        print(f'ERROR Reference geo transform array {dataset_ref} differs'
+              f'from secondary CSLC geo transform array {dataset_sec}')
+        return
+
```
seongsujeong, line 71: Would it be necessary to compare the projection?

seongsujeong (commented):
@vbrancat Nice work. I've left couple of comments here.

src/compass/utils/validate_cslc.py
vbrancat: As a note, I think I cannot apply this comment. See: https://stackoverflow.com/questions/40513144/open-geotiff-with-gdal-produces-attributeerror-exit

src/compass/utils/validate_cslc.py
vbrancat: That is a good idea :D

seongsujeong (approved):
Looks good to me

src/compass/utils/validate_cslc.py
```diff
@@ -0,0 +1,151 @@
...
+from osgeo import gdal
+
+
+def cmd_line_parser():
+    """
+    Command line parser
+    """
+
+    parser = argparse.ArgumentParser(description="""
+                                     Validate reference and generated (secondary) S1 CSLC products""",
```
yunjunz, line 15:
```suggestion
    parser = argparse.ArgumentParser(description="Validate reference and generated (secondary) S1 CSLC products",
```

To remove the long whitespaces.

yunjunz (approved):
Just one minor style suggestion below. The rest looks great to me!

PR #36, opened 2022-06-09
add `version.py` as the single-source version history

yunjunz:
This PR adds a `version.py` to store all the release history in one single source file, to access the `compass.__version__` variable and to pass it to the `setup.cfg` file.


seongsujeong (approved):
Looks good to me. Nice work!

PR #42, opened 2022-07-28
Dockerfile for S1 CSLC point release

seongsujeong:
This modification on Dockerfile is for point release of S1 CSLC software.
- the version of s1-reader has changed from `0.1.0` to `0.1.2`
- A fix that s1-reader gets installed outside of the desirable conda environment `COMPASS`
- The conda environment `COMPASS` in the container will be automatically activated when the user gets into the container by overriding the entry point to `/bin/bash`
- Addressing the WORKDIR issue suggested by @jshimada47

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,8 +5,8 @@
 FROM ubuntu:22.04
 
 LABEL author="OPERA ADT" \
-      description="s1 cslc 0.1.0 IF release" \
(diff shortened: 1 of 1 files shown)


rtburns-jpl (approved):
Looks good, nice work!

vbrancat (approved):

jshimada47 (commented):
Looks good, thanks!

PR #43, opened 2022-07-29
Place holder for model-based LUTs

vbrancat:
This PR introduces a place holder for the model-based LUTs we are going to be using in the geocoded burst workflow. 
At the moment, the s1-reader provides the computation of bistatic delay and total doppler.

Specifically, this PR introduces:
1. `utils/lut.py` function to compute a total slant range and azimuth LUT as a sum of individual LUTs
2. Expose spacings of LUTs to geocoded and radar runconfig/schemas
3. Introduce a place holder for slant range and azimuth LUTs in `s1_geocode_slc.py` and `s1_resample.py`. The overall LUTs should be subsequently fed to the corresponding core modules (`geocodeSlc` and `resample`) in the workflows


vbrancat:
@LiangJYu do you have time to have a look at it and check whether this is in line with the `geocodeSlc` implementation with LUTs?

seongsujeong (commented):
Looks good overall. I've added few minor suggestions.

src/compass/defaults/s1_cslc_radar.yaml
```diff
@@ -48,6 +48,9 @@ runconfig:
            resample:
                lines_per_block: 1000
                flatten: False
+           lut:
+               x_spacing: 500
+               y_spacing: 50
```
LiangJYu, line 53: Add same documenting comments found the geo defaults.

src/compass/schemas/s1_cslc_geo.yaml
```diff
@@ -91,6 +93,12 @@ range_split_spectrum_options:
   # Bandwidth of the high-band sub-image
   high_band_bandwidth: num(min=0, required=False)
 
+lut_options:
+  # LUT spacing in x/slant range direction
+  x_spacing: int(min=1, required=False)
+  # LUT spacing in y/azimuth direction
+  y_spacing: int(min=1, required=False)
```
LiangJYu, line 100:
```suggestion
  # LUT spacing in x/slant range direction in meters
  x_spacing: number(min=1, required=False)
  # LUT spacing in y/azimuth direction in seconds
  y_spacing: number(min=1, required=False)
```
In line with @seongsujeong's recommendation to use scientific notation. Also add units to documenting comments.

 Also applicable to radar schema.

src/compass/utils/lut.py
```diff
@@ -0,0 +1,22 @@
...
+    to be applied during burst geocoding
+
+    Parameters
+    ----------
+    burst: Sentinel1BurstSlc
+        S1-A/B burst object
+    xstep: int
+        LUT spacing (in pixels) along x/slant range
+    ystep: int
+        LUT spacing (in pixels) along x/azimuth
```
LiangJYu, line 13:
```suggestion
        LUT spacing (in pixels) along x/slant range in meters
    ystep: int
        LUT spacing (in pixels) along x/azimuth in seconds
```
Added units to be consistent with implementation in `isce3::geocode::geocodeSlc`.

src/compass/utils/lut.py
```diff
@@ -0,0 +1,22 @@
+def compute_lut(burst, xstep=500, ystep=50):
```
LiangJYu, line 1:
```suggestion
def compute_geocoding_correction_lut(burst, xstep=500, ystep=50):
```
Verbosity for clarity.

src/compass/s1_geocode_slc.py
```diff
@@ -62,6 +63,11 @@ def run(cfg):
     native_doppler = burst.doppler.lut2d
     orbit = burst.orbit
 
+    # Get range and azimuth LUTs
+    rg_lut, az_lut = compute_lut(burst,
```
LiangJYu, line 67:
```suggestion
    rg_correction_lut, az_correction_lut = compute_lut(burst,
```
Verbosity for clarity.

src/compass/s1_resample.py
```diff
@@ -11,6 +11,7 @@
 from osgeo import gdal
 
 from compass.utils.helpers import get_module_name
+from compass.utils.lut import compute_lut
```
LiangJYu, line 14: From my understanding, I'm not sure resample is the right place for the LUT2d correction to occur in the radar workflow. Based on the `isce3::geocode::geocodeSlc` [implementation](https://github-fn.jpl.nasa.gov/isce-3/isce/pull/1089). I think the corresponding location for these corrections should eventually be in [`s1_geocode`](https://github.com/opera-adt/COMPASS/pull/48/files). I'm assuming the radar workflow will eventually be geocoded?

Another issue with putting the corrections in resample, is how will these corrections be applied to the reference datasets, since only secondary datasets get resampled?

src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -58,6 +58,11 @@ runconfig:
               low_band_bandwidth: 10000000
               # High-band bandwidth in Hz
               high_band_bandwidth: 10000000
```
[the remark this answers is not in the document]
vbrancat: This comment has been addressed in the topo layers PR.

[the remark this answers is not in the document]
yunjunz: Scientific notation is the old way, isn't?

The old way is more readable to me.

src/compass/schemas/s1_cslc_geo.yaml
vbrancat: Just making sure before applying this change. At the moment, the only corrections we have in the `s1-reader` can be computed by setting the spacing parameters in terms of number of pixels (see https://github.com/opera-adt/s1-reader/blob/71e2c9e589f9e010b4952a574fcda83806561339/src/s1reader/s1_burst_slc.py#L401). For the schema/runconfig, do we want to use meters and second? I guess inside `lut.py`, we will need to fetch the number of pixels by dividing the spacing allocated in the runconfig by the spacing of the burst in slant range and azimuth directions. We can get the range_pixel_spacing from the Sentinel1BurstSlc object but not the azimuth spacing (which either should be hard-coded or extracted at center scene?). What do you recommend @LiangJYu @seongsujeong ? I would prefer to keep it in terms of number of pixels to avoid unnecessary conversions (especially getting the spacing in the azimuth direction is not trivial)

src/compass/s1_resample.py
vbrancat: Good point. One solution would be to remove the corrections from the radar workflow at the moment. You are correct, at a certain point we would want to geocode the burst on a geographical grid. However, for doing that, we will likely be using the `isce3.geocode.geocode_CFloat32()`. The core module does not support the use of correction LUTs at the moment. One solution would be to use `isce3.geocode.geocodeSlc()` and turn off the azimuth carrier resampling. Anyhow, I will remove the LUT computation from the radar workflow until we make a decision

vbrancat:
@LiangJYu do you perhaps have time to take another look at this?

src/compass/defaults/s1_cslc_radar.yaml
```diff
@@ -48,6 +48,9 @@ runconfig:
            resample:
                lines_per_block: 1000
                flatten: False
+           correction_luts:
+               x_spacing: 500
+               y_spacing: 50
```
LiangJYu, line 53:
```suggestion
               # LUT spacing in x/slant range direction in meters
               x_spacing: 500
               # LUT spacing in y/azimuth direction in seconds
               y_spacing: 50
```
Adding documentation.

src/compass/schemas/s1_cslc_radar.yaml
```diff
@@ -73,6 +73,12 @@ resample_options:
    # Flatten resampled SLC
    flatten: bool(required=False)
 
+lut_options:
+  # LUT spacing in x/slant range direction
+  x_spacing: int(min=1, required=False)
+  # LUT spacing in y/azimuth direction
```
LiangJYu, line 79:
```suggestion
  # LUT spacing in x/slant range direction in meters
  x_spacing: int(min=1, required=False)
  # LUT spacing in y/azimuth direction in seconds
```
Adding units

LiangJYu (commented):
Couple more comments. There are a couple comments I have not marked as "resolved" that need addressing. They're all documenting comment related.

vbrancat:
@LiangJYu @seongsujeong could you guys have a look at this PR earliest at your convenience?

LiangJYu (approved):
LGTM

src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -58,6 +58,11 @@ runconfig:
...
+          correction_luts:
+              # LUT spacing in x/slant range direction in meters
+              x_spacing: 500
+              # LUT spacing in y/azimuth direction in seconds
+              y_spacing: 50
```
seongsujeong, line 65: Not sure if 50 **seconds** of spacing in azimuth direction makes sense. Maybe spacing in meters?

src/compass/defaults/s1_cslc_radar.yaml
```diff
@@ -48,6 +48,11 @@ runconfig:
            resample:
                lines_per_block: 1000
                flatten: False
+           correction_luts:
+               # LUT spacing in x/slant range direction in meters
+               x_spacing: 500
+               # LUT spacing in y/azimuth direction in seconds
+               y_spacing: 50
```
seongsujeong, line 55: Same as above.

src/compass/utils/lut.py
```diff
@@ -0,0 +1,29 @@
+def compute_geocoding_correction_luts(burst, xstep=500, ystep=50):
```
seongsujeong, line 1: I do not think it needs to be addressed in this PR (because it would need a quite of work), but we might need to have an option to provide azimuth / range starting times by adding the arguments for these. That was necessary especially when I was defining the correction grids based on ETAD product.
If the starting time(s) is not provided, then we can bring them from `burst`

seongsujeong (commented):
@vbrancat I think this PR is almost ready to be merged. I've left few comments below.

src/compass/schemas/s1_cslc_geo.yaml
seongsujeong: I think we can use whichever makes more sense to calculate the correction, as long as we clearly mention about what that is. In case of FM mismatch rate, I stick with what ETAD ATBD is using for correction grid. (i.e. long time - short time).

vbrancat: That is interesting. So ETAD has seconds both in range and azimuth directions? I guess, and @LiangJYu can correct me if I am wrong, for the GSLC core module, we have used seconds for the LUT in azimuth and meters for the one along slant range?

src/compass/defaults/s1_cslc_geo.yaml
vbrancat: Your comment made me notice an inconsistency. So far, the only two corrections that we have (e.g. bistatic delay and geometrical + steering Doppler), they do require the spacings along azimuth and slant range expressed in terms of number of pixels. See https://github.com/opera-adt/s1-reader/blob/4000bc7a31d79e740212ab31495ca28ad0e2d426/src/s1reader/s1_burst_slc.py#L424 . To me, it makes sense to have it expressed in meters/seconds. How do we solve this? Shall we solve this in the s1-reader and change the steps parameters to support meters and seconds or we do the conversion internally? What do you think @LiangJYu ?

src/compass/utils/lut.py
vbrancat: Do we need to add this in the runconfig/schema? Would it be fine if the azimuth FM rate mismatch PR would take care of this?

src/compass/schemas/s1_cslc_geo.yaml
seongsujeong: > So ETAD has seconds both in range and azimuth directions?

Yes.

>I guess, and @LiangJYu can correct me if I am wrong, for the GSLC core module, we have used seconds for the LUT in azimuth and meters for the one along slant range?


That's my understanding about ISCE3. We use meters when defining range grid. See [here](https://github.com/opera-adt/s1-reader/blob/4000bc7a31d79e740212ab31495ca28ad0e2d426/src/s1reader/s1_burst_slc.py#L201-L209) and [here](https://github.com/opera-adt/s1-reader/blob/4000bc7a31d79e740212ab31495ca28ad0e2d426/src/s1reader/s1_reader.py#L569)
It would be great if @LiangJYu can confirm this.

LiangJYu: > > So ETAD has seconds both in range and azimuth directions?
> 
> Yes.
> 
> > I guess, and @LiangJYu can correct me if I am wrong, for the GSLC core module, we have used seconds for the LUT in azimuth and meters for the one along slant range?
> 
> That's my understanding about ISCE3. We use meters when defining range grid. See [here](https://github.com/opera-adt/s1-reader/blob/4000bc7a31d79e740212ab31495ca28ad0e2d426/src/s1reader/s1_burst_slc.py#L201-L209) and [here](https://github.com/opera-adt/s1-reader/blob/4000bc7a31d79e740212ab31495ca28ad0e2d426/src/s1reader/s1_reader.py#L569) It would be great if @LiangJYu can confirm this.

Confirming ISCE3 core geocodeSlc has az in seconds and slant range in meters.

src/compass/defaults/s1_cslc_geo.yaml
LiangJYu: Will these correction LUTs only be used by the core ISCE3 modules? If so then I think we should alter `burst.bistatic_delay` and `burst.geometrical_and_steering_doppler` to return LUT2d that support meters and seconds.

If there are other processes that expect these LUT2ds in pixels, we leave s1_reader untouched and change `compass.utils.lut.compute_geocoding_correction_luts` to support meters and seconds.

What do you all think?

src/compass/utils/lut.py
seongsujeong: > Do we need to add this in the runconfig/schema? 

Depends on how we define the correction grid on COMPASS. If we are going to define the correction grid **without** the starting timing for COMPASS production, then we wouldn't need to add them in runconfig.

Maybe adding the starting time would be more relevant to `s1-reader` than to `COMPASS`. I think adding the flexible starting times would be more relevant features when we are comparing our calculation with another source of computation (e.g. ETAD)

I will definitely include more freedom in timing when adding FM rate mismatch on `s1-reader`, but not sure if I am going to touch other corrections (e.g. `bistatic_delay`, `geometrical_and_steering_doppler`) because I have limited understanding to those.

src/compass/utils/lut.py
```diff
@@ -0,0 +1,29 @@
...
+    burst: Sentinel1BurstSlc
+        S1-A/B burst object
+    xstep: int
+        LUT spacing (in pixels) along x/slant range in meters
+    ystep: int
+        LUT spacing (in pixels) along x/azimuth in seconds
+
+    Returns
+    -------
+    rg_lut: ndarray
```
LiangJYu, line 17:
```suggestion
    rg_lut: isce.core.LUT2d
```

src/compass/utils/lut.py
```diff
@@ -0,0 +1,29 @@
...
+        2D array containing sum of range corrections
+    az_lut: ndarray
+        2D array containg sum of azimuth corrections
+    '''
+
+    az_bistatic = burst.bistatic_delay(xstep=xstep, ystep=ystep)
+    _, _, rg_doppler = burst.geometrical_and_steering_doppler(xstep=xstep, ystep=ystep)
+
+    rg_lut = rg_doppler
+    az_lut = az_bistatic.data
```
LiangJYu, line 27:
```suggestion
    az_lut = burst.bistatic_delay(xstep=xstep, ystep=ystep)
    rg_lut = burst.doppler_induced_range_shiftxstep=xstep, ystep=ystep)
```
More direct; the reassignment of burst corrections computation seem redundant? If the variable names were meant to convey info, I think additonal comments would better suited.
Range correction now a LUT2d instead of numpy array so it's something `isce.geocode.geocode_slc` can immediately use.

src/compass/utils/lut.py
vbrancat: So, I would suggest that we hold off on that :)

src/compass/defaults/s1_cslc_geo.yaml
vbrancat: Well, the primary user would be the `geocodeSlc` core module. However, providing a spacing in seconds in the runconfig/schema seems a bit counter intuitive to me. I am not sure which is the right approach. Shall we go ahead and modify the two corrections in the `s1-reader` to support meters and seconds? Are those two corrections used somewhere else (e.g. RTC)?

src/compass/utils/lut.py
vbrancat: Well, this is only one element of the look-up tables we will need for the `isce.geocode.geocode_slc`, so the reassignment is not redundant as later on we will have several elements adding up.

Is there a way to go from a numpy.array to a LUT2d using some ISCE3 functionalities?

src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -57,6 +57,13 @@ runconfig:
               low_band_bandwidth: 1e7
               # High-band bandwidth in Hz
               high_band_bandwidth: 1e7
+
+          correction_luts:
+              # LUT spacing in range direction in meters
+              range_spacing: 120
+              # LUT spacing in azimuth direction in meters
```
seongsujeong, line 64:
```suggestion
              # LUT spacing in azimuth direction in seconds
```

src/compass/defaults/s1_cslc_radar.yaml
```diff
@@ -68,6 +68,11 @@ runconfig:
...
```
seongsujeong, line 75: Is the LUT in radargrid? Then I would suggest to put more reasonable values here. Maybe the same as `s1_cslc_geo.yaml`
```suggestion
           correction_luts:
               # LUT spacing in x/slant range direction in meters
               x_spacing: 120
               # LUT spacing in y/azimuth direction in seconds
               y_spacing: 0.028
```

src/compass/schemas/s1_cslc_geo.yaml
```diff
@@ -93,6 +95,12 @@ range_split_spectrum_options:
   # Bandwidth of the high-band sub-image
   high_band_bandwidth: num(min=0, required=False)
 
+lut_options:
+  # LUT spacing in range direction in meters
+  range_spacing: num(min=1, required=False)
+  # LUT spacing in azimuth direction in meters
```
seongsujeong, line 101:
```suggestion
  # LUT spacing in azimuth direction in seconds
```

src/compass/schemas/s1_cslc_geo.yaml
```diff
@@ -93,6 +95,12 @@ range_split_spectrum_options:
...
+  azimuth_spacing: num(min=1, required=False)
```
seongsujeong, line 102: Not familiar with the schema, but `min=1` means the minimum value(s), or minumum **number** of the value? if the former, I would suggest to decrease it further in accordance with the default value in `s1_cslc_geo.yaml` (i.e. 0.028 seconds)

src/compass/schemas/s1_cslc_radar.yaml
```diff
@@ -85,6 +85,12 @@ resample_options:
    # Flatten resampled SLC
    flatten: bool(required=False)
 
+lut_options:
+  # LUT spacing in x/slant range direction in meters
+  x_spacing: int(min=1, required=False)
+  # LUT spacing in y/azimuth direction in seconds
+  y_spacing: int(min=1, required=False)
```
seongsujeong, line 92: Same as above

seongsujeong (commented):
Thanks for lots of discussion about the LUTs. The comments I've just left are nothing but minor issues. See the below for the details.

src/compass/schemas/s1_cslc_geo.yaml
vbrancat: Goof catch. I will decrease it to zero so that we can plug in only positive numbers

seongsujeong (approved):
LGTM. Thanks for addressing the comments I've posted.
Please make sure to resolve the codacy issue before merging.

vbrancat:
@seongsujeong FYI, I cannot resolve the codacy issues at the moment because they are related to the fact that we are not using the `rg_lut` and `az_lut` variables in the code. This will need to be integrated when we release a new version of ISCE3 with conda

PR #47, opened 2022-08-11
setup: access version w/o __init__ to avoid circular imports

yunjunz:
This is the same as https://github.com/opera-adt/s1-reader/pull/56


vbrancat (approved):
Thank you, @yunjunz

LiangJYu (approved):
LGTM

seongsujeong (approved):
Looks good to me

PR #48, opened 2022-08-18
Compute and geocode topo layers in geocode CSLC workflow

vbrancat:
This PR adds the generation and the geocoding of various topo layers (incidence angle, local incidence angle, azimuth angle, layover shadow mask) in the geocode CSLC workflow.

The generation of the topo layers reuses (and modifies) the `s1_rdr2geo.py` module which generates the topo layers in radar coordinate at full-resolution. The PR introduces a new script `s1_geocode_topo.py` to geocode the produced layers. 

Note that we have 2 use cases:
1. We want to generate full-resolution range/Doppler topo layers to use in the computation of range, azimuth model-based LUT corrections (i.e., solid Earth tides)
2. PCM might generate the topo layers once for a stack of bursts and allocated them in a separate file (TBC).

Therefore, we should be able to switch on/off the topo layers we want to generate.

TO DO:
1. Extend the radar CSLC workflow to geocode the topo layers: need to modify `runconfig.py` to accomodate the generation of the `geogrid`. Decided to address this in another PR.
2. Confirm that `localPsi` in ISCE3 corresponds to the ISCE2 azimuth angle definition


src/compass/s1_geocode_topo.py
```diff
@@ -0,0 +1,144 @@
...
+    output_rasters = [
+        isce3.io.Raster(f'{output_path}/{fname}.geo',
+                        geo_grid.width, geo_grid.length, 1, gdal.GDT_Float32,
+                        output_format)
+        if enabled else None
+        for fname, enabled in topo_layers.items()
+    ]
+    # Filter None from list
+    input_rasters = [raster for raster in input_rasters if raster is not None]
+    output_rasters = [raster for raster in output_rasters if raster is not None]
```
LiangJYu, line 104:
```suggestion
    input_rasters = [
        isce3.io.Raster(f'{input_path}/{fname}.rdr')
        for fname, enabled in topo_layers.items() if enabled]        
    output_rasters = [
        isce3.io.Raster(f'{output_path}/{fname}.geo',
                        geo_grid.width, geo_grid.length, 1, gdal.GDT_Float32,
                        output_format)
        for fname, enabled in topo_layers.items() if enabled]
```
No need to remove `None` by moving `enabled` check after `for` loop.

LiangJYu (approved):
LGTM. Just one small nit.

vbrancat:
@yunjunz and @seongsujeong Does one of you have time to review this PR? Merci :)

seongsujeong (commented):
Looks good to me. My only suggestion is to use scientific notation for `low_band_bandwidth` and `high_band_bandwidth` as suggested in another PR [here](https://github.com/opera-adt/COMPASS/pull/43/files#r945989364). That would also be help users to recognize them as floating point numbers, not integers.

vbrancat:
@seongsujeong your comment should have been addressed by my last commit. I have used the scientific notation to allocate the low and high band bandwidth default values for the split-spectrum. Please, earliest at your convenience, can you have another look at the PR?

vbrancat:
@seongsujeong and @yunjunz would any of you have time to give an extra check to this PR?

vbrancat:
@seongsujeong or @yunjunz can one of you review this PR? Thanks.

scottstanie:
At the suggestion of Heresh, I think we'll want to avoid using "topo" as much as possible in name, and rename the script to something like `s1_geocode_imaging_geometry.py`, or (if you think that's too long), `s1_geocode_metadata.py`

vbrancat:
@scottstanie thanks. I changed the nomenclature from `topo` to `metadata` as per one of your suggestions

seongsujeong (approved):
LGTM

PR #49, opened 2022-08-19
Passing more dependencies

sssangha:
Without these, the program fails to run

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,6 @@ gdal>=3
 #journal      # as of Mar 2022, journal from conda does not support python3.9; since it is included during isce3 installation above, comment this out temporarily.
 ruamel.yaml
 yamale
+packaging
+backoff
+requests


requirements.txt
```diff
@@ -5,3 +5,6 @@ gdal>=3
 #journal      # as of Mar 2022, journal from conda does not support python3.9; since it is included during isce3 installation above, comment this out temporarily.
 ruamel.yaml
 yamale
+packaging
```
vbrancat, line 8: @yunjunz I am not sure where this dependency comes from. The other two: `backoff` comes from `isce3` (the `stage_dem.py` script requires it), `requests` comes from the Sentinel-1 reader from the script where we download the orbits. Wouldn't it make sense to allocate these two dependencies there?

yunjunz: I am not sure why we need `packaging`. Could you @sssangha elaborate?

As for the other two, they should not be added to the `compass` repo, as none of them are not directly used. 

I believe all of us got the `backoff` module import error while installing isce3 through conda. @vbrancat is it possible for you to add it to the isce3 conda recipe?

@sssangha could you please issue a PR to add `requests` to `s1reader` instead?

LiangJYu: @sssangha piggybacking off @yunjunz comment above, `packaging` should also be included in `s1reader`.

Also did you install isce3 from conda or did you build it from scratch?

IIRC either @rtburns-jpl or @gmgunter do the isce3 conda upload. I'll talk to them on how include `backoff` as a part of the recipe.

yunjunz:
Close this PR since all 3 new dependencies have been added to the upstream repos:
+ `backoff` has been added to the isce3 conda-forge recipe (https://github.com/conda-forge/isce3-feedstock/pull/29)
+ `packaging` and `requests` have been added to the `s1reader` (https://github.com/opera-adt/s1-reader/pull/68).

PR #52, opened 2022-10-04
switch to `s1reader.s1_orbit.get_orbit_file_from_dir()`

vbrancat:
This should fix a bug in COMPASS created by the latest merge in the `s1-reader` repo. Since the function `get_orbit_file_from_list` now accepts a directory and not a list of orbit files, I am wondering whether we should change the schema to provide a path to an orbit directory instead of a list of orbit files


yunjunz (approved):
LGTM, thank you @vbrancat.

src/compass/utils/runconfig.py
```diff
@@ -161,9 +161,9 @@ def runconfig_to_bursts(cfg: SimpleNamespace) -> list[Sentinel1BurstSlc]:
     # extract given SAFE zips to find bursts identified in cfg.burst_id
     for safe_file in cfg.input_file_group.safe_file_path:
         # get orbit file
-        orbit_path = get_orbit_file_from_list(
+        orbit_path = get_orbit_file_from_dir(
```
LiangJYu, line 164:
```suggestion
# get orbit file from directory of first orbit file
```
nit - modify comment above for clarity

LiangJYu (approved):
LGTM. Thanks for catching+fixing. Nitpick below.

PR #53, opened 2022-10-11
Geocoded burst stack processor (cont)

scottstanie:
Creating this PR mostly from Virginia's #18 . I used it last week for 3-4 areas, so it's pretty functional. I came across several things as I was processing, notably
- the dates in the dataframe/code were integers, but were trying to be filtered using the pandas `.str.contains`
- The dates passed to `--exclude-dates` were actually being included, so I refactored the filter function to allow both exclusions or inclusions (but fixed the logic so `--exclude-dates` does exclude)

I think I've addressed several of the comments made on #18, so I was just trying to help fix up the final changes others can more easily work on the same code, without merging multiple forked branches.


scottstanie:
@LiangJYu I believe I implemented all the suggestions you had in comments here https://github.com/opera-adt/COMPASS/pull/18 . The only one I didn't do was moving all pruning calls into the data frame generation, since after profiling, the pandas filtering was actually very quick even with a few thousand bursts.
So if there aren't objections, I'll merge this in shortly.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+        (left, bottom, right, top)
+    """
+    x_left = x_spac * np.floor(geogrid.start_x / x_spac)
+    x_right = x_spac * np.ceil((geogrid.start_x + x_spac * geogrid.width) / x_spac)
+    y_bottom = y_spac * np.floor((geogrid.start_y + y_spac * geogrid.length) / y_spac)
+    y_top = y_spac * np.ceil(geogrid.start_y / y_spac)
+    return (x_left, y_bottom, x_right, y_top)
+
+
+
```
LiangJYu, line 171:
```suggestion
```
Maintaining consistent 2 spaces between functions.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+        column identification for 'data' (e.g. 'burst_id')
+    id_list: list
+        List of elements to consider when pruning.
+        If exclude_items is False (default), then all elements in `data`
+            will be kept *except for* those in `id_list`.
+        If exclude_items is True, the items in `id_list` will be removed from `data`.
+    exclude_items: bool
+        If True, the items in `id_list` will be removed from `data`.
+        If False, all elements in `data` will be kept *except for* those in `id_list`.
+    Returns:
```
LiangJYu, line 191:
```suggestion

    Parameters:
    ----------
    data: pandas.DataFrame
        dataframe that needs to be pruned
    id_col: str
        column identification for 'data' (e.g. 'burst_id')
    id_list: list
        List of elements to consider when pruning.
        If exclude_items is False (default), then all elements in `data`
            will be kept *except for* those in `id_list`.
        If exclude_items is True, the items in `id_list` will be removed from `data`.
    exclude_items: bool
        If True, the items in `id_list` will be removed from `data`.
        If False, all elements in `data` will be kept *except for* those in `id_list`.
        
    Returns:
```
Empty lines for readability.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+    return df
+
+
+def get_common_burst_ids(data):
+    '''
+    Get list of burst IDs common among all processed dates
+    Parameters:
+    ----------
+    data: pandas.DataFrame
+      Dataframe containing info for stitching (e.g. burst IDs)
```
LiangJYu, line 207:
```suggestion

    Parameters:
    ----------
    data: pandas.DataFrame
      Dataframe containing info for stitching (e.g. burst IDs)
      
```
Extra lines for readability.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+    Parameters:
+    ----------
+    zip_file_list: list
+        List of zip files to filter
+    start_date: str
+        Start date in YYYYMMDD format
+    end_date: str
+        End date in YYYYMMDD format
+    exclude_dates: list
+        List of dates to exclude
```
LiangJYu, line 316:
```suggestion

    Parameters:
    ----------
    zip_file_list: list
        List of zip files to filter
    start_date: str
        Start date in YYYYMMDD format
    end_date: str
        End date in YYYYMMDD format
    exclude_dates: list
        List of dates to exclude
        
```
Extra lines for readability.

LiangJYu (approved):
LGTM. My comments are all formatting related.

vbrancat (commented):
@scottstanie thanks for all the improvements :). I only have minor comments at this stage. I am planning to test the PR pretty soon and will let you know if anything comes up.

One "big observation" I have is the following. Since we have a burst map, I was wondering whether for the stack processor, we want to use the info in the burst map to populate the runconfig (e.g. epsg, geogrid). We are following this direction for CSLC processing. We will soon have a place holder in the schema for the burst map and extrapolate epsg and geogrid bbox for each burst from the burst map itself. What do you think? We should not address this point in this PR but it would be good to keep in mind.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+                        help='File path to DEM to use for processing.')
+    parser.add_argument('-o', '--orbit-dir', default=None,
+                        help='Directory with orbit files. If None, downloads orbit files')
+    parser.add_argument('-w', '--working-dir', dest='work_dir', default='stack',
+                        help='Directory to store intermediate and final results')
+    parser.add_argument('-sd', '--start-date', help='Start date of the stack to process')
+    parser.add_argument('-ed', '--end-date', help='End date of the stack to process')
+    parser.add_argument('-b', '--burst-id', nargs='+', default=None,
+                        help='List of burst IDs to process. If None, all the burst IDs '
+                             'in the reference date are processed. (default: None)')
```
hfattahi, line 38: what is the reference date? I don't see it in the list of input arguments. Actually I don't know why would we need reference date at all?

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+                        help='Date to be excluded from stack processing (format: YYYYMMDD)')
+    parser.add_argument('-p', '--pol', dest='pol', nargs='+', default='co-pol',
+                        help='Polarization to process: dual-pol, co-pol, cross-pol '
+                             ' (default: co-pol).')
+    parser.add_argument('-x', '--x-spac', type=float, default=5,
+                        help='Spacing in meters of geocoded CSLC along X-direction.')
+    parser.add_argument('-y', '--y-spac', type=float, default=10,
+                        help='Spacing in meters of geocoded CSLC along Y-direction.')
+    parser.add_argument('--bbox', nargs=4, type=float, default=None, 
+                        metavar=('lonmin', 'latmin', 'lonmax', 'latmax'),
```
hfattahi, line 49: Any reason why allowing only lat-lon bbox and not other coordinate systems like UTM etc? if we allow UTM, then we need another argument to specify the bbox-epsg.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+                        help='(Optional) Bounding box of the geocoded stack'
+                             ' in lon/lat degrees. ')
+    parser.add_argument('-e', '--epsg', type=int, default=None,
```
hfattahi, line 52: I think we should have at least an option to use our burst database for the extracting epsg and geocoded bounding box. This way we will get to also test that database.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+                        help='EPSG projection code for output geocoded bursts')
+    parser.add_argument('-f', '--flatten', type=bool, default=True,
+                        help='If True, enables flattening (default: True)')
+    parser.add_argument('-ss', '--range-split-spectrum',
+                        dest='is_split_spectrum', type=bool, default=False,
+                        help='If True, enables split-spectrum (default: False)')
+    parser.add_argument('-lb', '--low-band', type=float, default=0.0,
+                        help='Low sub-band bandwidth in Hz (default: 0.0)')
+    parser.add_argument('-hb', '--high-band', type=float, default=0.0,
```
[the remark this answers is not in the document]
hfattahi: Yes they are needed for down stream workflows. We should generate them only once (one date) and not for every date.

src/compass/utils/helpers.py
```diff
@@ -154,3 +156,56 @@ def check_dem(dem_path: str):
...
+        Tuple containing the bounding box coordinates in UTM (meters)
+        (left, bottom, right, top)
+    """
+    coords = np.array(poly.exterior.coords)
+    xys = _convert_to_utm(coords, epsg)
+    return geometry.Polygon(xys)
+
+
+def _convert_to_utm(points_lonlat, epsg):
+    """Convert a list of points from lon/lat to UTM coordinates."""
```
hfattahi, line 205:
```suggestion
    """Convert a list of points from lon/lat (in degrees) to UTM coordinates."""
```

hfattahi (commented):
Thanks @scottstanie for restoring this PR. I put in few minor comment. I think it will be useful for us to have an option to use the burst ID database. Maybe it is not bad to allow the current behavior if the burst database is not provided. 

Another point is do we want to generate the geocoded bursts only over the common ID among all dates? I think if we force processing over only common busrst IDs among all dates, then we could potentially exclude a lot of busrsts to process. An alternative would be to loop over the zip files and process each burst that statisfies the input constraints (Burst ID list, or bbox). This way we may end up with different number of coregsitered SLCs for different IDs.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+    parser.add_argument('-p', '--pol', dest='pol', nargs='+', default='co-pol',
+                        help='Polarization to process: dual-pol, co-pol, cross-pol '
+                             ' (default: co-pol).')
+    parser.add_argument('-x', '--x-spac', type=float, default=5,
+                        help='Spacing in meters of geocoded CSLC along X-direction.')
+    parser.add_argument('-y', '--y-spac', type=float, default=10,
+                        help='Spacing in meters of geocoded CSLC along Y-direction.')
+    parser.add_argument('--bbox', nargs=4, type=float, default=None, 
+                        metavar=('lonmin', 'latmin', 'lonmax', 'latmax'),
+                        help='(Optional) Bounding box of the geocoded stack'
```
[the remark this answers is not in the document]
scottstanie: Just checking your thoughts on this-
I added the `ArgumentDefaultsHelpFormatter` class:
```python
    parser = argparse.ArgumentParser(
        description='S1-A/B geocoded CSLC stack processor.',
        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
which means that the defaults are auto printed out, e.g.

```
-s SLC_DIR, --slc-dir SLC_DIR
                        Directory containing the S1-A/B SLCs (zip files) (default: None)
```

Also, in the --help text of argparse:

```
$ s1_geo_stack.py --help
usage: s1_geo_stack.py [-h] -s SLC_DIR -d DEM_FILE [-o ORBIT_DIR] [-w WORK_DIR] ...
...
```
The ones that have `required=True` have no square brackets around them (-s, -d) but the rest do have. brackets.


But, if you think this isn't clear enough, I found a thing that nicely separates the required options clearly without having it at the end of the string every time: https://stackoverflow.com/a/41747010/

which looks like this:

```
 s1_geo_stack.py --help
usage: s1_geo_stack.py [-h] -s SLC_DIR -d DEM_FILE [-o ORBIT_DIR] [-w WORK_DIR] [-sd START_DATE] [-ed END_DATE] [-b BURST_ID [BURST_ID ...]] [-exd EXCLUDE_DATES [EXCLUDE_DATES ...]] [-p POL [POL ...]] [-dx X_SPAC] [-dy Y_SPAC]
                       [--bbox lonmin latmin lonmax latmax] [--bbox-epsg BBOX_EPSG] [-e EPSG] [-f FLATTEN] [-ss IS_SPLIT_SPECTRUM] [-lb LOW_BAND] [-hb HIGH_BAND]

S1-A/B geocoded CSLC stack processor.

required arguments:
  -s SLC_DIR, --slc-dir SLC_DIR
                        Directory containing the S1-A/B SLCs (zip files) (default: None)
  -d DEM_FILE, --dem-file DEM_FILE
                        File path to a GDAL-readable DEM to use for processing. (default: None)

optional arguments:
  -o ORBIT_DIR, --orbit-dir ORBIT_DIR
                        Directory with orbit files. If None, downloads orbit files (default: None)
```

src/compass/s1_geo_stack.py
scottstanie: Removed the bad reference to reference date

src/compass/s1_geo_stack.py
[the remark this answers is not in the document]
vbrancat: I like that. That would give a clear separation between required and optional arguments. If it is not too much work, it would be awesome to add it.

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,490 @@
...
+    end_time = time.time()
+    print('Elapsed time (min):', (end_time - start_time) / 60.0)
+
+
+def main():
+    """Create the command line interface and run the script."""
+    # Run main script
+    args = create_parser()
+
+    main(args.slc_dir, args.dem_file, args.burst_id, args.start_date,
```
vbrancat, line 482: I think this function is supposed to be called `run`

vbrancat:
Another nit-picking consideration. Shall we consider renaming the `s1_geo_stack.py` function? In the main folder, we opted for explicitly spell the word `geocode`.  May be something like `s1_geocoded_cslc_stack_processor`? It might be too long but something like that would be helpful.

src/compass/s1_geo_stack.py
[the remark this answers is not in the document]
scottstanie: Ok I've added a flag to generate the metadata for each burst, only on the first date

src/compass/s1_geo_stack.py
```diff
@@ -0,0 +1,468 @@
...
+    i_subswath = [1, 2, 3]
+
+    for zip_file in zip_files:
+        orbit_path = get_orbit_file_from_dir(zip_file, orbit_dir)
+
+        for subswath in i_subswath:
+            ref_bursts = load_bursts(zip_file, orbit_path, subswath)
+            for burst in ref_bursts:
+                if epsg is None:
+                    epsg = get_point_epsg(burst.center.y,
```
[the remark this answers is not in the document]
scottstanie: I've added the use of the database instead of calculating it.

One note- I didn't know whether to have the default behavior be

1. every burst gets the EPSG from the database. So if a frame straddles a boundary, it will have some from each zone.
2. pick the first EPSG we see (which was how it was implemented), then translate the boundaries from the database to match.

I've opted for 1 for now, since you can specify one EPSG with the command line arg, but you can't easily say "use the database for everything"

vbrancat:
@scottstanie For next time, we do implicitly require that, for each PR, "at least" two people grant approvals. This is fine for now, but please, make sure to follow this rule for your next PR.

PR #56, opened 2022-10-18
Bug in creating metadata layers

vbrancat:
This PR fix a bug in the creation of the metadata layers. Layers need to be geocoded only when the execution of `rdr2geo` is requested.


scottstanie (approved):

LiangJYu (approved):
LGTM

PR #60, opened 2022-11-07
make common burst filtering optional in stack processor

scottstanie:
This addresses the comment from @hfattahi that I missed in the stack processor change before: https://github.com/opera-adt/COMPASS/pull/53#pullrequestreview-1148555147
> do we want to generate the geocoded bursts only over the common ID among all dates? I think if we force processing over only common busrst IDs among all dates, then we could potentially exclude a lot of bursts to process. 

This makes that common burst filter optional with the `--common-bursts-only` flag.


vbrancat (approved):
LGTM, thanks, and sorry for getting back to it so late. This is a very nice and much-needed functionality !!! Make sure to resolve conflicts before merging.

vbrancat:
@LiangJYu would you have time to get to this PR? It is a very easy one. Merci

LiangJYu (approved):
LGTM

PR #64, opened 2022-11-21
EAP correction into COMPASS workflow

seongsujeong:
This PR integrates elevation antenna pattern correction (EAP correction) into COMPASS workflow. As discussed, the functionality will only be added into `s1_geocode_slc.py`
Also the EAP correction will take whenever is necessary i.e. No user control by the flag in runconfig. 
The field `eap_correction_by_opera` in the metadata will indicate if the EAP correction is applied by the SAS.

### DIFF
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -11,12 +11,13 @@ import journal
 import numpy as np
 from osgeo import gdal
 
+from compass.utils.elevation_antenna_pattern import apply_eap_correction
 from compass.utils.geo_metadata import GeoCslcMetadata
 from compass.utils.geo_runconfig import GeoRunConfig
 from compass.utils.helpers import get_module_name
 from compass.utils.range_split_spectrum import range_split_spectrum
 from compass.utils.yaml_argparse import YamlArgparse
-
+from s1reader.s1_reader import is_eap_correction_necessary
 
 def run(cfg):
     '''
@@ -74,6 +75,20 @@ def run(cfg):
     else:
         temp_slc_path = f'{scratch_path}/{burst_id}_{pol}_temp.vrt'
         burst.slc_to_vrt_file(temp_slc_path)
+
+        # Check if EAP correction is necessary; Apply the correction if yes
+        check_eap = is_eap_correction_necessary(burst.ipf_version)
+        if check_eap.phase_correction:
+            temp_slc_path_in = temp_slc_path
+            temp_slc_path_corrected = temp_slc_path_in.replace('_temp.vrt',
+                                                               '_corrected_temp.rdr')
+            burst.slc_to_vrt_file(temp_slc_path_in)
+            apply_eap_correction(burst,
+                                 temp_slc_path_in, 
+                                 temp_slc_path_corrected,
+                                 check_eap)
+            temp_slc_path = temp_slc_path_corrected
+
         rdr_burst_raster = isce3.io.Raster(temp_slc_path)
 
     # Generate output geocoded burst raster
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -92,7 +92,7 @@ def run(cfg: GeoRunConfig):
             burst.slc_to_vrt_file(temp_slc_path)
             rdr_burst_raster = isce3.io.Raster(temp_slc_path)
 
-            # Check if EAP correction is necessary; Apply the correction if True
+            # Check if EAP correction is necessary; Apply the correction if it is True
             check_eap = is_eap_correction_necessary(burst.ipf_version)
             if check_eap.phase_correction:
                 temp_slc_path_in = temp_slc_path
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -135,7 +135,7 @@ def run(cfg: GeoRunConfig):
         geotransform = [geo_grid.start_x, geo_grid.spacing_x, 0,
                         geo_grid.start_y, 0, geo_grid.spacing_y]
         geo_burst_raster.set_geotransform(geotransform)
-        geo_burst_raster.set_epsg(epsg)
+        geo_burst_raster.set_epsg(geo_grid.epsg)
         del geo_burst_raster
         del dem_raster # modified in geocodeSlc
 
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -100,7 +100,7 @@ def run(cfg: GeoRunConfig):
                                                                    '_corrected_temp.rdr')
                 burst.slc_to_vrt_file(temp_slc_path_in)
                 apply_eap_correction(burst,
-                                    temp_slc_path_in, 
+                                    temp_slc_path_in,
                                     temp_slc_path_corrected,
                                     check_eap)
                 temp_slc_path = temp_slc_path_corrected
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -45,6 +45,7 @@ def run(cfg):
     iters = cfg.geo2rdr_params.numiter
     blocksize = cfg.geo2rdr_params.lines_per_block
     flatten = cfg.geocoding_params.flatten
+    output_epsg = cfg.geocoding_params.output_epsg
 
     # process one burst only
     burst = cfg.bursts[0]
@@ -103,7 +104,7 @@ def run(cfg):
     geotransform = [geo_grid.start_x, geo_grid.spacing_x, 0,
                     geo_grid.start_y, 0, geo_grid.spacing_y]
     geo_burst_raster.set_geotransform(geotransform)
-    geo_burst_raster.set_epsg(epsg)
+    geo_burst_raster.set_epsg(output_epsg)
     del geo_burst_raster
 
     # Save burst metadata
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -82,7 +82,7 @@ def run(cfg: GeoRunConfig):
             if cfg.rdr2geo_params.geocode_metadata_layers:
                s1_geocode_metadata.run(cfg, fetch_from_scratch=True)
 
-        
+
         #Load the input burst SLC
         temp_slc_path = f'{scratch_path}/{id_pol}_temp.vrt'
         burst.slc_to_vrt_file(temp_slc_path)
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -92,7 +92,7 @@ def run(cfg: GeoRunConfig):
             burst.slc_to_vrt_file(temp_slc_path)
             rdr_burst_raster = isce3.io.Raster(temp_slc_path)
 
-            # Check if EAP correction is necessary; Apply the correction if yes
+            # Check if EAP correction is necessary; Apply the correction if True
             check_eap = is_eap_correction_necessary(burst.ipf_version)
             if check_eap.phase_correction:
                 temp_slc_path_in = temp_slc_path
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -82,28 +82,33 @@ def run(cfg: GeoRunConfig):
             if cfg.rdr2geo_params.geocode_metadata_layers:
                s1_geocode_metadata.run(cfg, fetch_from_scratch=True)
 
+        
+        #Load the input burst SLC
+        temp_slc_path = f'{scratch_path}/{id_pol}_temp.vrt'
+        burst.slc_to_vrt_file(temp_slc_path)
+
+        # Check if EAP correction is necessary
+        check_eap = is_eap_correction_necessary(burst.ipf_version)
+        if check_eap.phase_correction:
+            temp_slc_path_corrected = temp_slc_path.replace('_temp.vrt',
+                                                            '_corrected_temp.rdr')
+            apply_eap_correction(burst,
+                                 temp_slc_path,
+                                 temp_slc_path_corrected,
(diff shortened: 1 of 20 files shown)


src/compass/utils/elevation_antenna_pattern.py
```diff
@@ -0,0 +1,69 @@
...
+        EAP correction are necessary
+
+    Return:
+    -------
+    path_burst_corrected:
+        Path to the GDAL-compatible raster after the correction
+
+    '''
+
+    #TODO Indicate into the metadata to indicate if EAP correction is applied by OEPRA SAS
```
vbrancat, line 33: When the PR gets merged (or also now), submit an issue about adding the application of EAP to the CSLC metadata. So we do not lose track on it

src/compass/utils/elevation_antenna_pattern.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+    if not check_eap.phase_correction:
+        print('Antenna pattern correction is not necessary.')
+        return None
```
vbrancat, line 38: Do we need to return None. I would prefer checking if EAP is necessary in `s1_geocode_slc.py`. If the flag `check_eap` is `True` you execute the `elevation_antenna_pattern.py`. This eliminate the need for an extra argument in the function and is a bit more transparent toward the users

src/compass/utils/elevation_antenna_pattern.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Retrieve the EAP correction in range
+    vec_eap_line = burst.eap_compensation_lut
+
+    # Remove magnitute component when we don't need to correct it
```
vbrancat, line 43:
```suggestion
    # Remove magnitude component when we don't need to correct it
```

src/compass/utils/elevation_antenna_pattern.py
```diff
@@ -0,0 +1,69 @@
...
+
+    if check_eap.phase_correction and not check_eap.magnitude_correction:
+        vec_eap_line /= np.abs(vec_eap_line)
+
+    
+
```
vbrancat, line 49:
```suggestion
```

No more than 2 lines between lines of code

src/compass/utils/geo_metadata.py
```diff
@@ -196,7 +206,8 @@ def from_file(cls, file_path: str, fmt: str):
                    meta_dict['range_window_type'],
                    meta_dict['range_window_coefficient'], cfg, geogrid,
                    meta_dict['nodata'], meta_dict['input_data_ipf_version'],
-                   meta_dict['isce3_version'])
+                   meta_dict['isce3_version'],
+                   meta_dict['eap_correction_by_opera'])
```
vbrancat, line 210: I would simply write `antenna_patter_correction`. We are re-crafting the output product allocation as a HDF5 file, in the attribute description we will make sure to indicate that this is a Boolean flag indicating if the correction has been applied or not. Since you are including this in the metadata, no need to submit an issue (as previously indicated in my comment) but do remove the comment ;)

src/compass/s1_geocode_slc.py
```diff
@@ -89,6 +92,19 @@ def run(cfg: GeoRunConfig):
             burst.slc_to_vrt_file(temp_slc_path)
             rdr_burst_raster = isce3.io.Raster(temp_slc_path)
 
+            # Check if EAP correction is necessary; Apply the correction if yes
```
vbrancat, line 95: May be "if it is True"?

vbrancat (commented):
Thank you @seongsujeong for this PR. I have left some comments. I was not able to test the PR. I think I have found some very old Sentinel data which require the application of EAP. 
However, for some reason, it seems that I am unable to load the AUX CAL files when I am operating in a conda environment. Below, the error I incurred into:

```
>>> bursts = load_bursts(file, orbit_path, 1)
  File "<stdin>", line 1, in <module>
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.10/site-packages/s1reader/s1_reader.py", line 757, in load_bursts
    bursts = _burst_from_zip(path, id_str, orbit_path, flag_apply_eap)
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.10/site-packages/s1reader/s1_reader.py", line 818, in _burst_from_zip
    bursts = burst_from_xml(f_annotation, orbit_path, f_tiff, iw2_f_annotation, z_file.open,
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.10/site-packages/s1reader/s1_reader.py", line 515, in burst_from_xml
    path_aux_cal = get_path_aux_cal(path_aux_cals, annotation_path)
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.10/site-packages/s1reader/s1_reader.py", line 326, in get_path_aux_cal
    raise ValueError( 'Cannot find AUX_CAL files from directory: '
ValueError: Cannot find AUX_CAL files from directory: /mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.10/site-packages/s1reader/data/aux_cal
```

Not sure if I am doing anything wrong. I will try to manually copy the files in the directory and give the code a try

vbrancat:
I tested the code on a S1A SAFE file needing EAP and the code runs successfully. Unfortunately, at this round, I do not have much time to check the InSAR phase and whether the correction has been correctly applied. I would recommend to submit an issue where we remind ourselves that we have to perform some sort of validation of the EAP with historical Sentinel data. Can you do it for me?

As I had a look at the code, for me, it makes sense to move the EAP check before here https://github.com/seongsujeong/COMPASS/blob/87b48a810e478fd7eddbc401acef13fdd802017b/src/compass/s1_geocode_slc.py#L85. In this way, we can generate the temporary corrected burst before applying the split-spectrum (no need to include the correction inside the split-spectrum code) and if the split-spectrum is not required we can directly load the temporary corrected EAP burst into an `isce3.io.Raster` object. What do you think? Any opinion @LiangJYu ?

src/compass/utils/elevation_antenna_pattern.py
seongsujeong: Forgot to remove this TODO. The metadata writeout is implemented in `geo_metadata.py`.
I will remove this line.

src/compass/utils/elevation_antenna_pattern.py
seongsujeong: Checking the EAP correction take places on this code and `s1_geocode_slc.py`. I agree that that the check here would be redundant. I will remove the lines here.

src/compass/utils/geo_metadata.py
seongsujeong: I was also thought about `antenna_patter_correction`, but felt that it would be necessary to indicate if the correction is applied by ESA (i.e. IPF >= 2.43) or OPERA, so that we can track back if there are any issues regarding the EAP correction.

What do you think about using the field name `antenna_patter_correction`, and putting the string `ESA` or `OPERA`?

src/compass/s1_geocode_slc.py
seongsujeong: Yes, I think your suggestion looks more clear. The comment was revised.

seongsujeong:
Thanks @vbrancat for taking a time to review this PR. I've addressed your comments, and left a question [here](https://github.com/opera-adt/COMPASS/pull/64#discussion_r1030635096) regarding the metadata.

I've just submitted an issue to remind us with the validation test (for EAP).

>As I had a look at the code, for me, it makes sense to move the EAP check before here https://github.com/seongsujeong/COMPASS/blob/87b48a810e478fd7eddbc401acef13fdd802017b/src/compass/s1_geocode_slc.py#L85. In this way, we can generate the temporary corrected burst before applying the split-spectrum (no need to include the correction inside the split-spectrum code) and if the split-spectrum is not required we can directly load the temporary corrected EAP burst into an `isce3.io.Raster` object. What do you think?

I think that's a good idea. If that's the case, the function `range_split_spectrum()` will need to take in the corrected burst, but I don't think the change would be tricky.

vbrancat:
> Thanks @vbrancat for taking a time to review this PR. I've addressed your comments, and left a question [here](https://github.com/opera-adt/COMPASS/pull/64#discussion_r1030635096) regarding the metadata.
> 
> I've just submitted an issue to remind us with the validation test (for EAP).
> 
> > As I had a look at the code, for me, it makes sense to move the EAP check before here https://github.com/seongsujeong/COMPASS/blob/87b48a810e478fd7eddbc401acef13fdd802017b/src/compass/s1_geocode_slc.py#L85. In this way, we can generate the temporary corrected burst before applying the split-spectrum (no need to include the correction inside the split-spectrum code) and if the split-spectrum is not required we can directly load the temporary corrected EAP burst into an `isce3.io.Raster` object. What do you think?
> 
> I think that's a good idea. If that's the case, the function `range_split_spectrum()` will need to take in the corrected burst, but I don't think the change would be tricky.

Once you have solved the issue of giving to the split-spectrum the corrected EAP burst directly, this PR should be good to go :). Thanks for the quick PR :D

seongsujeong:
@vbrancat Thanks for the feedback. Do you have some preference about the metadata issue in the comment here?
https://github.com/opera-adt/COMPASS/pull/64#discussion_r1030635096

> What do you think about using the field `name antenna_patter_correction`, and putting the string `ESA` or `OPERA`?

vbrancat:
> 

@LiangJYu can take care of it. He can make sure to have a field in the metadata to cover when the antenna pattern correction. I would not hold this PR for the metadata. From my side is good to go (we will make sure to do some testing after Beta delivery). @LiangJYu it would be good if you can also look at it and make sure to integrate a metadata field for the antenna pattern correction

src/compass/s1_geocode_slc.py
```diff
@@ -79,16 +82,34 @@ def run(cfg: GeoRunConfig):
             if cfg.rdr2geo_params.geocode_metadata_layers:
                s1_geocode_metadata.run(cfg, fetch_from_scratch=True)
 
+
+        #Load the input burst SLC
```
LiangJYu, line 86:
```suggestion
        # Load the input burst SLC
```
nit - extra space

LiangJYu (approved):
LGTM aside from unresolved metadata comment. My comments are mostly formatting related.

vbrancat (approved):
LGTM, thanks @seongsujeong

hfattahi:
Thanks @seongsujeong for this PR. As we discussed, let's do a test on real data to make sure things work properly before merging the PR. We need to make an interferogram between SLCs processed with IPF 2.36 and an IPF above and see the impact of the EAP correction. The expected results should look like figure 2 in [this document](https://sentinels.copernicus.eu/documents/247904/1653440/Sentinel-1-IPF_EAP_Phase_correction.pdf/d8767e29-ae1d-4e1b-92c2-22c10648f81f?t=1437645629000) .

src/compass/utils/geo_metadata.py
seongsujeong: Replacing the field name to Liang's suggestion.

src/compass/s1_geocode_slc.py
```diff
@@ -79,16 +82,34 @@ def run(cfg: GeoRunConfig):
...
+        temp_slc_path = f'{scratch_path}/{id_pol}_temp.vrt'
+        burst.slc_to_vrt_file(temp_slc_path)
+
+        # Check if EAP correction is necessary
+        check_eap = is_eap_correction_necessary(burst.ipf_version)
+        if check_eap.phase_correction:
+            temp_slc_path_corrected = temp_slc_path.replace('_temp.vrt',
+                                                            '_corrected_temp.rdr')
+            apply_eap_correction(burst,
```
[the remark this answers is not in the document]
seongsujeong: @LiangJYu Good point. Maybe we should pass `isce3.io.Raster` objects, rather than path to the rasters as `str`. Same for `range_split_spectrum` that follows the EAP correction.
For that I would need to figure out how to have read/write access to the data in ISCE3 raster. Please create an issue for that, and let's discuss about the operations there.

seongsujeong:
Merging the PR to COMPASS. Thank you for your help @vbrancat @LiangJYu !

seongsujeong:
> Thanks @seongsujeong for this PR. As we discussed, let's do a test on real data to make sure things work properly before merging the PR. We need to make an interferogram between SLCs processed with IPF 2.36 and an IPF above and see the impact of the EAP correction. The expected results should look like figure 2 in [this document](https://sentinels.copernicus.eu/documents/247904/1653440/Sentinel-1-IPF_EAP_Phase_correction.pdf/d8767e29-ae1d-4e1b-92c2-22c10648f81f?t=1437645629000) .

Thanks @hfattahi for the suggestion. Here is the screenshot of the mosaicked interferograms from geoburst SLCs. The first screenshot is the result with EAP correction turned off, and the second data is when the correction turned on.

Screen Shot 2022-12-02 at 11 43 10
Screen Shot 2022-12-02 at 11 47 30

Below are more information about the processed data:
reference SLC: `S1A_IW_SLC__1SSV_20150207T014953_20150207T015023_004511_005899_9C44`, IPF = 2.36 (EAP was NOT applied)
secondary SLC: `S1A_IW_SLC__1SDV_20150327T014954_20150327T015021_005211_00694B_4E66`, IPF = 2.43 (EAP was  applied)

seongsujeong:
Just to note, I'm pushing couple of fixes here that make the CSLC workflow works as expected

- After the PR [here](https://github.com/opera-adt/s1-reader/pull/77/files), `Sentinel1BurstSlc.burst_id` is not a `str` but a class  I've edited few lines on COMPASS to handle this change.

- We were using the EPSG of DEM. I've changed it to take the number from burst DB (to be precisely, geogrid generated from the burst DB)

hfattahi (approved):
Looks great. Thanks @seongsujeong for adding the evidence.

Issue #73, opened 2022-12-07
[change the main dataset name]: 


hfattahi:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

the main dataset is named "complex_backscatter" and that seems too long and not fully represnetative of the content. The developer team agrees to change to "slc"

### Describe the feature request

rename "complex_backscatter" to "slc" in HDF5

yunjunz:
Close this issue as it has been implemented in #74 by @LiangJYu.

PR #74, opened 2022-12-07
rename main dataset name from complex_backscatter to SLC

LiangJYu:
Addresses #73 by changing main dataset name from  `complex_backscatter` to `SLC`


seongsujeong:
LGTM, as long as there is not any sort of convention to use lowercase for the dataset name.

seongsujeong (approved):

LiangJYu:
> LGTM, as long as there is not any sort of convention to use lowercase for the dataset name.

Good point. NISAR has `SLC`, `RSLC`, and `GSLC` but that doesn't necessarily make it right...

yunjunz (approved):
LGTM.

PR #75, opened 2022-12-14
update hdf5 metadata structure and content

LiangJYu:
The PR adds:
- writing azimuth and slant range correction LUT2ds to `/correction` group in HDF5
- moves EAP to `/correction` group
- change organization of groups and datasets to align with RTC HDF5


vbrancat (changes requested):
@LiangJYu as discussed yesterday through Slack, it is preferred to save the correction separately. Can you apply this change? For each correction, we might need to separate the LUT in range and azimuth. The product structure will be something like:

root
  - SLC
    - polarization
  - corrections
      - doppler
           - rg_lut
      - solid_earth_tides
           - rg_lut
           - az_lut
      - slant_range
      - azimuth_time

src/compass/utils/h5_helpers.py
```diff
@@ -446,3 +427,80 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+    check_eap = is_eap_correction_necessary(burst.ipf_version)
+    if check_eap.phase_correction:
+        eap_group = correction_group.require_group('elevation_antenna_pattern_correction')
+        eap = burst.burst_eap
+        add_dataset_and_attrs(eap_group, 'sampling_frequency',
+                              eap.freq_sampling,
+                              {'description': 'range sampling frequency',
+                               'units': 'Hz'})
+        add_dataset_and_attrs(eap_group, 'eta_start',
+                              eap.eta_start.strftime(TIME_STR_FMT),
```
seongsujeong, line 486: Same as above.

src/compass/utils/h5_helpers.py
```diff
@@ -446,3 +427,80 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+                              {'description': '?',
```
seongsujeong, line 487: `eta_start` is sensing start time. `eta` came from the greek letter in the reference document [here](https://sentinels.copernicus.eu/documents/247904/1653440/Sentinel-1-IPF_EAP_Phase_correction.pdf/d8767e29-ae1d-4e1b-92c2-22c10648f81f?t=1437645629000) (section 5.3)

src/compass/utils/h5_helpers.py
```diff
@@ -446,3 +427,80 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+        add_dataset_and_attrs(eap_group, 'theta_sub', eap.theta_sub,
+                              {'description': 'elevation angle',
+                               'units': 'radians'})
+        add_dataset_and_attrs(eap_group, 'azimuth_time',
+                              eap.azimuth_time.strftime(TIME_STR_FMT),
+                              {'description': '?',
+                               'format': 'YYYY-MM-DD HH:MM:SS.6f'})
+        add_dataset_and_attrs(eap_group, 'ascending_node_time',
+                              eap.ascending_node_time.strftime(TIME_STR_FMT),
+                              {'description': '?',
```
seongsujeong, line 504:
```suggestion
                              {'description': 'Ascending node crossing time (ANX)',
```

src/compass/utils/h5_helpers.py
```diff
@@ -446,3 +427,80 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+                               'units': 'seconds'})
+        add_dataset_and_attrs(eap_group, 'tau_sub', eap.tau_sub,
+                              {'description': 'slant range time',
+                               'units': 'seconds'})
+        add_dataset_and_attrs(eap_group, 'theta_sub', eap.theta_sub,
+                              {'description': 'elevation angle',
+                               'units': 'radians'})
+        add_dataset_and_attrs(eap_group, 'azimuth_time',
+                              eap.azimuth_time.strftime(TIME_STR_FMT),
+                              {'description': '?',
```
seongsujeong, line 500: realized that the member `BurstEAP.azimuth_time` is redundant (i.e. same value as `BurstEAP.eta_start`.
I'll remove this member in the follow-up PR to `s1-reader`.

seongsujeong (commented):
Nice work. Here I leave few comments as below.

src/compass/utils/h5_helpers.py
```diff
@@ -243,7 +254,7 @@ def add_dataset_and_attrs(group, name, value, attr_dict):
     add_dataset_and_attrs(id_proc_group, 'product_level', '2', {})
     add_dataset_and_attrs(id_proc_group, 'product_type', 'CSLC_S1', {})
     add_dataset_and_attrs(id_proc_group, 'processing_datetime',
-                          datetime.now().strftime(time_str_fmt),
+                          datetime.now().strftime('%Y-%m-%d %H:%M:%S'),
```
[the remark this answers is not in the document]
LiangJYu: I chose not to use `TIME_STR_FMT = '%Y-%m-%d %H:%M:%S.%f'` because I don't think this piece of information requires detail of fractional seconds. I can change if you and others disagree.

Since the format string ` '%Y-%m-%d %H:%M:%S` is single use, I saw no need to declare it elsewhere.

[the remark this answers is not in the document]
seongsujeong: Thanks for the clarifying about your intention. Then it seems like most of my comments about the formatting can be solved by fixing `TIME_STR_FMT = '%Y-%m-%d %H:%M:%S'` into `TIME_STR_FMT = '%Y-%m-%d %H:%M:%S.%f'`

src/compass/utils/h5_helpers.py
```diff
@@ -3,10 +3,16 @@
...
 import numpy as np
 from osgeo import osr
+from s1reader.s1_reader import is_eap_correction_necessary
 
 import compass
+from compass.utils.lut import compute_geocoding_correction_luts
 from compass.utils.raster_polygon import get_boundary_polygon
 
+
+TIME_STR_FMT = '%Y-%m-%d %H:%M:%S'
```
seongsujeong, line 13: based on your response, I think you would have meant `'%Y-%m-%d %H:%M:%S.%f'`?

LiangJYu: Agreed. I will make that change. Will push that along with overhaul of group/dataset hierarchy.

LiangJYu:
FYI - In the last couple commits, I left some questions as comments in this PR's code.

src/compass/utils/h5_helpers.py
```diff
@@ -193,256 +202,269 @@ def init_geocoded_dataset(geocoded_group, dataset_name, geo_grid, dtype,
...
+
+    # doppler correction LUTs
+    desc = ' correction as a function of slant range and azimuth time'
+    doppler_correction_items = [
+        meta('azimuth_LUT', az_lut.data, f'azimuth {desc}',
+             {'units': 'seconds'}),
+        meta('slant_range_LUT', rg_lut.data, f'slant_range {desc}',
+             {'units': 'seconds'}),
+    ]
+    doppler_correction_group = dst_h5.require_group('/CSLC/grids/corrections/doppler')
```
vbrancat, line 440: This contribution is the sum of the geometrical and steering Doppler. We should specify it in the variable name. Rename it to something like `geometry_steering_doppler`?

vbrancat: Do the datasets have any descriptions? If not, we should add one for completeness

vbrancat (commented):
Thank you @LiangJYu, this is a good start.

Some consideration below.

1. My understanding from @hfattahi is that we want CSLC to be similar to RTC. Therefore, we need to add `/science/SENTINEL1/` as main group, similarly to what RTC does
2. SLC group should be changed in CSLC.
3. At the moment, the code fails at line 
    ```
File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.10/site-packages/compass/utils/h5_helpers.py", line 340, in geo_burst_metadata_to_hdf5
    if 'orbit' in dst_h5['/CSLC/metadata']:
As the CSLC group is not found and the metadata group does not exists in the code.
```

File: src/compass/utils/h5_helpers.py [hunk 8]
```diff
@@ -193,256 +202,269 @@ def init_geocoded_dataset(geocoded_group, dataset_name, geo_grid, dtype,
...
+        # NOTE: in NISAR, the value has to be in UPPERCASE or lowercase?
+        meta('look_direction', 'Right', 'Look direction can be left or right'),
+        meta('orbit_pass_direction', burst.orbit_direction,
+             'Orbit direction can be ascending or descending'),
+        # NOTE: using the same date format as `s1_reader.as_datetime()`
+        meta('zero_doppler_start_time', burst.sensing_start.strftime(TIME_STR_FMT),
+             'Azimuth start time of product'),
+        meta('zero_doppler_end_time', burst.sensing_stop.strftime(TIME_STR_FMT),
+            'Azimuth stop time of product'),
+        meta('list_of_frequencies', ['A'],
```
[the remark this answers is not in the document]
LiangJYu: It's copy of RTC. I can remove if you like.

src/compass/utils/h5_helpers.py
```diff
@@ -193,256 +202,269 @@ def init_geocoded_dataset(geocoded_group, dataset_name, geo_grid, dtype,
...
+             'Input noise file used'),
+        meta('dem_source', os.path.basename(cfg.dem), 'DEM source description')
+    ]
+    input_group = dst_h5.require_group('/CSLC/metadata/processingInformation/inputs')
+    for meta_item in input_items:
+        add_dataset_and_attrs(input_group, meta_item)
+
+    # algorithm items
+    algorithm_items = [
+        meta('dem_interpolation', 'BIQUINTIC', 'DEM interpolation method'),
```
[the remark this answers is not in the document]
LiangJYu: :laughing: Good point. No need for angry, aggressive interpolation.

src/compass/utils/h5_helpers.py
LiangJYu: Adopted your suggested name.

They do have descriptions.

vbrancat:
Thank you, @LiangJYu. it looks much better right now. A couple of minor fixes and it should be ready to go:

1.  `/science/SENTINEL1/CSLC/corrections/zero_doppler_time` has the shape of the `slant_range` vector not of the azimuth. In addition, the `corrections` group should have the slant range and zero doppler time spacing of the correction LUTs
2. In `/science/SENTINEL1/CSLC/grids/`, we should remove spacing of the burst in radar coordinates and add those in geocoded coordinates (like `x_spacing`, `y_spacing`). I would also move the `center_frequency` and `slant_range_bandwidth` to the metadata section where we collect the info on the input burst
3. before, we had a section of the metadata where we had a bunch of info related to the input burst named `s1ab_burst_metadata`. Can we have it back and put center_frequency` and `slant_range_bandwidth` there?

After that, we are good to go. Awesome work :)

LiangJYu:
> 2. In `/science/SENTINEL1/CSLC/grids/`, we should remove spacing of the burst in radar coordinates and add those in geocoded coordinates (like `x_spacing`, `y_spacing`). I would also move the `center_frequency` and `slant_range_bandwidth` to the metadata section where we collect the info on the input burst

Can you point to where I'm using radar coordinates. I don't think I'm using radar coordinates anywhere in the metadata. In `/science/SENTINEL1/CSLC/grids/`, the CSLC axis are in [geo coordinates](https://github.com/opera-adt/COMPASS/blob/442616577bd7f1a7469f1b43c778cb9c5c7832dd/src/compass/utils/h5_helpers.py#L41-L58).

LiangJYu:
> 3. before, we had a section of the metadata where we had a bunch of info related to the input burst named `s1ab_burst_metadata`. Can we have it back and put center_frequency`and`slant_range_bandwidth` there?

Restore [other datasets to `s1ab_burst_metadata`](https://github.com/opera-adt/COMPASS/blob/3ed0702e7a79a73ee749bc2aceea339676773032/src/compass/utils/h5_helpers.py#L264-L320) too?

LiangJYu:
> 4. The corrections have contributions in range and azimuth. At the moment, they are allocated as 2-D arrays. Not sure if it is best to separate the slant range and azimuth LUTs in a separate group or allocated them as 3D matrices

Like the following?
```
* corrections
  - doppler
    - azimuth (aka bistatic_delay)
    - range (aka geometry_steering_doppler)
```

vbrancat:
Answering to your questions in one comment.
2. Correct, you are not using radar data in that group. My point is that having metadata related to range/Doppler products in a group where we have geocoded data is confusing. I propose to restore the metadata group `s1ab_burst_metadata` and insert all this info in this group
3. Yes, let's restore `s1ab_burst_metadata`
4. I like your solution

hfattahi:
Thanks @LiangJYu  for the PR. I think this is very exciting. Based on a product that @vbrancat pointed me to here are some observations:

The bounding polygon does not seem correct to me. It seems to be in pixel while it should be geo-coordinates

bounding_polygon b'POLYGON ((17968 546, 15156 808, 7580 1514, 5992 1662, 2752 1964, 1369 2093, 1170 2112, 1156 2120, 1152 2125, 1144 2141, 1142 2152, 1161 2238, 1189 2315, 1880 4093, 1915 4114, 2111 4096, 2423 4067, 17945 2617, 18476 2567, 18824 2534, 18896 2527, 18904 2522, 18907 2517, 18121 745, 17973 549, 17968 546))'


I wonder if we should use lower case for ```burst_ID``` to be ```burst_id``` and the ```mission_ID``` to be ```mission_id```. 

For product version please make sure that a field in the run config exists for "product_version". The SAS should read that field and place in the product_version metadata. Maybe that is already in place? (Note that SDS will be responsible for populating the product_version).
 
shall we use ```epsg_code``` instead of ```projection```?
 
 The data type is complex64 (two float 32). I thought we are going with float16. Is that a separate PR?
 
 I suggest to change ```slc['science/SENTINEL1/CSLC/metadata/processing_information/algorithms/geocoding']``` to ```slc['science/SENTINEL1/CSLC/metadata/processing_information/algorithms/geocode_interpolator']``` or ```geocode_interpolation```
 
 This is great to keep track the input radar geometry here ```slc['science/SENTINEL1/CSLC/metadata/processing_information/s1_burst_metadata']```, would you please make sure all information to reconstruct the radarGridParameters exists? I don't see zerod Doppler start and stop time. I also don't see the stop_range. We need either stop time and stop range or width length of the radar grid to reconstruct the radar grid parameters. 
 
 Please change ```azimuth_steer_rate``` to ```azimuth_steering_rate```
 
 I think it will be really nice to be able to reconstruct the burst SLC object with metadata in the CSLC product. Basically here we could keep track of all these parameters: https://github.com/opera-adt/s1-reader/blob/main/src/s1reader/s1_burst_slc.py#L209  . We actually had this functionality when we had json format.

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +524,161 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('azimuth_last_azimuth_line', noise.azimuth_last_azimuth_line,
+                 'azimuth_last_azimuth_line'),
+            Meta('azimuth_last_range_sample', noise.azimuth_last_range_sample,
+                 'azimuth_last_range_sample'),
+            Meta('azimuth_line', noise.azimuth_line, 'azimuth_line'),
+            Meta('azimuth_lut', noise.azimuth_lut, 'azimuth_lut')
+        ]
+        noise_group = correction_group.require_group('noise')
+        for meta_item in noise_items:
+            add_dataset_and_attrs(noise_group, meta_item)
```
LiangJYu, line 684: @seongsujeong Can you please take a look at the description strings for these items? The descriptions I've used are insufficient. Thanks.

vbrancat (approved):
LGTM, thanks for having worked on it. It looks awesome :)

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
-                          geogrid.epsg,
-                          {'description': 'EPSG code identifying the coordinate system used for processing'})
-    add_dataset_and_attrs(processing_group, 'no_data_value', 'NaN',
-                          {'description': 'Value used when no data present'})
+        eap_items = [
+            Meta('sampling_frequency', eap.freq_sampling,
+                 'range sampling frequency', { 'units': 'Hz'}),
+            Meta('eta_start', eap.eta_start.strftime(TIME_STR_FMT),
+                 'Sensing start time', {'format': 'YYYY-MM-DD HH:MM:SS.6f'}),
+            Meta('tau_0', eap.tau_0, 'slant range time', {'units': 'seconds'}),
```
seongsujeong, line 635:
```suggestion
            Meta('tau_0', eap.tau_0, 'slant range time of the product', {'units': 'seconds'}),
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('tau_sub', eap.tau_sub, 'slant range time',
```
seongsujeong, line 636:
```suggestion
            Meta('tau_sub', eap.tau_sub, 'slant range time of AUX_CAL antenna pattern',
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+    if burst.burst_noise is not None:
+        noise = burst.burst_noise
+        noise_items = [
+            Meta('basename', noise.basename_nads, ''),
+            Meta('range_azimith_time',
+                 noise.range_azimith_time.strftime(TIME_STR_FMT),
+                 'Start time', {'format': 'YYYY-MM-DD HH:MM:SS.6f'}),
+            Meta('range_line', noise.range_line, 'Range line'),
+            Meta('range_pixel', noise.range_pixel, 'Range pixel array'),
+            Meta('range_lut', noise.range_lut, 'Range lookup table data'),
```
seongsujeong, line 677:
```suggestion
            Meta('range_lut', noise.range_lut, 'Range noise lookup table data'),
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('azimuth_first_azimuth_line', noise.azimuth_first_azimuth_line,
+                 'azimuth_first_azimuth_line'),
+            Meta('azimuth_first_range_sample', noise.azimuth_first_range_sample,
+                 'azimuth_first_range_sample'),
+            Meta('azimuth_last_azimuth_line', noise.azimuth_last_azimuth_line,
+                 'azimuth_last_azimuth_line'),
+            Meta('azimuth_last_range_sample', noise.azimuth_last_range_sample,
+                 'azimuth_last_range_sample'),
+            Meta('azimuth_line', noise.azimuth_line, 'azimuth_line'),
+            Meta('azimuth_lut', noise.azimuth_lut, 'azimuth_lut')
```
seongsujeong, line 687:
```suggestion
            Meta('azimuth_lut', noise.azimuth_lut, 'azimuth noise lookup table data')
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('range_lut', noise.range_lut, 'Range lookup table data'),
+            Meta('azimuth_first_azimuth_line', noise.azimuth_first_azimuth_line,
+                 'azimuth_first_azimuth_line'),
+            Meta('azimuth_first_range_sample', noise.azimuth_first_range_sample,
+                 'azimuth_first_range_sample'),
+            Meta('azimuth_last_azimuth_line', noise.azimuth_last_azimuth_line,
+                 'azimuth_last_azimuth_line'),
+            Meta('azimuth_last_range_sample', noise.azimuth_last_range_sample,
+                 'azimuth_last_range_sample'),
+            Meta('azimuth_line', noise.azimuth_line, 'azimuth_line'),
```
seongsujeong, line 686:
```suggestion
            Meta('azimuth_line', noise.azimuth_line, 'azimuth line index for noise LUT'),
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+    # write out noise metadata, if present
+    if burst.burst_noise is not None:
+        noise = burst.burst_noise
+        noise_items = [
+            Meta('basename', noise.basename_nads, ''),
+            Meta('range_azimith_time',
+                 noise.range_azimith_time.strftime(TIME_STR_FMT),
+                 'Start time', {'format': 'YYYY-MM-DD HH:MM:SS.6f'}),
+            Meta('range_line', noise.range_line, 'Range line'),
+            Meta('range_pixel', noise.range_pixel, 'Range pixel array'),
```
seongsujeong, line 676:
```suggestion
            Meta('range_pixel', noise.range_pixel, 'Range in pixel for LUT'),
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('range_lut', noise.range_lut, 'Range noise lookup table data'),
+            Meta('azimuth_first_azimuth_line', noise.azimuth_first_azimuth_line,
+                 'azimuth_first_azimuth_line'),
```
seongsujeong, line 679:
```suggestion
                 'First line of the burst in subswath. None if not available in annotation.'),
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('azimuth_first_range_sample', noise.azimuth_first_range_sample,
+                 'azimuth_first_range_sample'),
```
seongsujeong, line 681:
```suggestion
                 'First range sample of the burst. None if not available in annotation.'),
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('azimuth_last_azimuth_line', noise.azimuth_last_azimuth_line,
+                 'azimuth_last_azimuth_line'),
```
seongsujeong, line 683:
```suggestion
                 'Last line of the burst in subswatn. None if not available in annotation.'),
```

src/compass/utils/h5_helpers.py
```diff
@@ -320,129 +532,160 @@ def poly1d_to_h5(group, poly1d_name, poly1d):
...
+            Meta('azimuth_last_range_sample', noise.azimuth_last_range_sample,
+                 'azimuth_last_range_sample'),
```
seongsujeong, line 685:
```suggestion
                 'Last range of the burst. None if not available in annotation.'),
```

hfattahi (approved):
Thank you @LiangJYu for addressing my comments. With the caveat of this issue https://github.com/opera-adt/COMPASS/issues/76 the rest LGTM.

PR #77, opened 2022-12-22
Modify CSLC validation script

vbrancat:
The CSLC validation script performs a pixel-wise comparison of the difference between the real and imaginary parts of a reference and secondary products and fails when even one pixel is above a fixed threshold (1.0e-5).

This PR keeps the pixel-wise comparison but makes the validation script fail if the percentage of pixels for which the difference between the real and imaginary parts is above 0.1 %.

### DIFF
--- a/src/compass/utils/validate_cslc.py
+++ b/src/compass/utils/validate_cslc.py
@@ -135,10 +135,10 @@ def compare_cslc_products(file_ref, file_sec):
     for dtype_op, dtype_str in zip([np.real, np.imag],
                                    ['real', 'imag']):
         # compute difference pixel by pixel
-        diff = (dtype_op(ma_slc_ref) - dtype_op(ma_slc_sec)) / dtype_op(ma_slc_sec)
+        diff = (dtype_op(ma_slc_ref) - dtype_op(ma_slc_sec)) / dtype_op(ma_slc_ref)
 
         # count number of pixels that exceed threshold
-        n_diff_gt_threshold = np.count_non_zero(diff > pixel_diff_threshold)
+        n_diff_gt_threshold = np.count_nonzero(diff > pixel_diff_threshold)
 
         # compute percentage of pixels that fail
         percent_fail = n_diff_gt_threshold / tot_pixels_ref
@@ -148,9 +148,9 @@ def compare_cslc_products(file_ref, file_sec):
             dtype_fails.append(dtype_str)
 
     # check no fails occurred by check of
-    err_str = f'Percentage of pixels in the difference between reference and '
-              f'secondary products parts above the threshold {pixel_diff_threshold} '
-              f'is above {percent_fail_threshold*100}% for: '
+    err_str = f'Percentage of pixels in the difference between reference ' \
+              f'and secondary products parts above the threshold ' \
+              f'{pixel_diff_threshold} is above {percent_fail_threshold*100}% for: '
     assert len(dtype_fails) == 0, err_str + ','.join(dtype_fails)
 
 
--- a/src/compass/utils/validate_cslc.py
+++ b/src/compass/utils/validate_cslc.py
@@ -30,12 +30,10 @@ def _gdal_nfo_retrieve(path_h5):
     """
     Extract polarization, geotransform array, projection, and SLC from
     given HDF5
-
     Parameters
     ----------
     path_h5: str
         File path to CSLC HDF5 product
-
     Returns
     -------
     pol: str
@@ -69,7 +67,6 @@ def compare_cslc_products(file_ref, file_sec):
     '''
     Compare a reference and a newly generated
     (i.e., secondary) CSLC product
-
     Parameters
     ----------
     file_ref: str
@@ -110,12 +107,9 @@ def compare_cslc_products(file_ref, file_sec):
         return
 
     # Compare amplitude of reference and generated CSLC products
-    nan_ref = np.isnan(slc_ref)
-    nan_sec = np.isnan(slc_sec)
-    np.testing.assert_array_equal(nan_ref, nan_sec)
-
-    ma_slc_ref = np.ma.masked_array(slc_ref, mask=~nan_ref)
-    ma_slc_sec = np.ma.masked_array(slc_sec, mask=~nan_sec)
+    np.seterr(invalid='ignore')
+    diff_real = np.abs((slc_ref.real - slc_sec.real) / slc_sec.real)
+    diff_imag =  np.abs((slc_ref.imag - slc_sec.imag) / slc_sec.imag)
 
     # Compute total number of pixels different from nan from ref and sec
     tot_pixels_ref = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
@@ -124,45 +118,31 @@ def compare_cslc_products(file_ref, file_sec):
     # Check that total number of pixel is the same
     assert tot_pixels_ref == tot_pixels_sec
 
-    # Compute percentage of pixels in real/imaginary part above threshold
-    pixel_diff_threshold = 1e-5
-    percent_fail_threshold = 0.001
+    # Compute the number of pixels in real part above threshold
+    failed_pixels_real = np.count_nonzero(diff_real > 1.0e-5)
+    failed_pixels_imag = np.count_nonzero(diff_imag > 1.0e-5)
 
-    # container to capture all possible dtype failures
-    # tests passes if empty i.e. no failures were captured
-    dtype_fails = []
+    # Compute percentage of pixels in real and imaginary part above threshold
+    percentage_real = failed_pixels_real / tot_pixels_ref
+    percentage_imag = failed_pixels_imag / tot_pixels_ref
 
-    for dtype_op, dtype_str in zip([np.real, np.imag],
-                                   ['real', 'imag']):
-        # compute difference pixel by pixel
-        diff = (dtype_op(ma_slc_ref) - dtype_op(ma_slc_sec)) / dtype_op(ma_slc_ref)
+    # Check that percentage of pixels above threshold is lower than 0.1 %
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products real parts above the threshold 1.0e-5 is below 0.1 %')
+    assert percentage_real < 0.001
 
-        # count number of pixels that exceed threshold
-        n_diff_gt_threshold = np.count_nonzero(diff > pixel_diff_threshold)
-
-        # compute percentage of pixels that fail
-        percent_fail = n_diff_gt_threshold / tot_pixels_ref
-
-        # Check that percentage of pixels above percent_fail_threshold
-        if percent_fail > percent_fail_threshold:
-            dtype_fails.append(dtype_str)
-
-    # check no fails occurred by check of
-    err_str = f'Percentage of pixels in the difference between reference ' \
-              f'and secondary products parts above the threshold ' \
-              f'{pixel_diff_threshold} is above {percent_fail_threshold*100}% for: '
-    assert len(dtype_fails) == 0, err_str + ','.join(dtype_fails)
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products imaginary parts above the threshold 1.0e-5 is below 0.1 %')
+    assert percentage_imag < 0.001
 
 
 def _get_group_item_paths(h5py_group):
     '''
     Get paths for all datasets and groups nested within a h5py.Group
-
     Parameters
     ----------
     h5py_group: h5py.Group
         Group object where paths to objects within are to be retrieved.
-
     Returns
     -------
     paths: list[str]
@@ -184,7 +164,6 @@ def _get_group_item_paths(h5py_group):
 def compare_cslc_metadata(file_ref, file_sec):
     '''
     Compare reference and generated CSLC metadata
-
     Parameters
     ----------
     file_ref: str
--- a/src/compass/utils/validate_cslc.py
+++ b/src/compass/utils/validate_cslc.py
@@ -110,9 +110,12 @@ def compare_cslc_products(file_ref, file_sec):
         return
 
     # Compare amplitude of reference and generated CSLC products
-    np.seterr(invalid='ignore')
-    diff_real = np.abs((slc_ref.real - slc_sec.real) / slc_sec.real)
-    diff_imag =  np.abs((slc_ref.imag - slc_sec.imag) / slc_sec.imag)
+    nan_ref = np.isnan(slc_ref)
+    nan_sec = np.isnan(slc_sec)
+    np.testing.assert_array_equal(nan_ref, nan_sec)
+
+    ma_slc_ref = np.ma.masked_array(slc_ref, mask=~nan_ref)
+    ma_slc_sec = np.ma.masked_array(slc_sec, mask=~nan_sec)
 
     # Compute total number of pixels different from nan from ref and sec
     tot_pixels_ref = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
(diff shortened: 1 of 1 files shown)


src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,32 @@ def compare_cslc_products(file_ref, file_sec):
...
+    pixels_imag = np.count_nonzero(diff_imag > 1.0e-5)
+
+    # Compute percentage of pixels in real and imaginary part above threshold
+    percentage_real = pixels_real / tot_pixels_ref
+    percentage_imag = pixels_imag / tot_pixels_ref
+
+    # Check that percentage of pixels above threshold is lower than 0.1 %
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products real parts is below 0.1 %')
+    assert percentage_real < 0.1
```
seongsujeong, line 134:
```suggestion
    assert percentage_real < 0.001
```
0.1 would be actually 10%. 0.001 Would be correct one.

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,32 @@ def compare_cslc_products(file_ref, file_sec):
...
+
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products imaginary parts is below 0.1 %')
+    assert percentage_imag < 0.1
```
seongsujeong, line 138: Same as above

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,32 @@ def compare_cslc_products(file_ref, file_sec):
...
+    # Compute total number of pixels different from nan from ref and sec
+    tot_pixels_ref = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
+    tot_pixels_sec = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
+
+    # Check that total number of pixel is the same
+    assert tot_pixels_ref == tot_pixels_sec
+
+    # Compute the number of pixels in real part above threshold
+    pixels_real = np.count_nonzero(diff_real > 1.0e-5)
+    pixels_imag = np.count_nonzero(diff_imag > 1.0e-5)
```
seongsujeong, line 125:
```suggestion
    failed_pixels_real = np.count_nonzero(diff_real > 1.0e-5)
    failed_pixels_imag = np.count_nonzero(diff_imag > 1.0e-5)
```

nit pick - and clarity

seongsujeong (commented):
Thanks for the revision on the validation script @vbrancat. Here I leave some comments as below

src/compass/utils/validate_cslc.py
vbrancat: Good catch :)

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,32 @@ def compare_cslc_products(file_ref, file_sec):
...
+    failed_pixels_real = np.count_nonzero(diff_real > 1.0e-5)
+    failed_pixels_imag = np.count_nonzero(diff_imag > 1.0e-5)
+
+    # Compute percentage of pixels in real and imaginary part above threshold
+    percentage_real = failed_pixels_real / tot_pixels_ref
+    percentage_imag = failed_pixels_imag / tot_pixels_ref
+
+    # Check that percentage of pixels above threshold is lower than 0.1 %
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products real parts is below 0.1 %')
```
seongsujeong, line 133:
```suggestion
    print('Check that the percentage of pixels in the difference between reference'
          'and secondary products real parts above the threshold 1.0e-5 is below 0.1 %')
```

A bit more information about the threshold and pass/fail criteria

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,32 @@ def compare_cslc_products(file_ref, file_sec):
...
+    assert percentage_real < 0.001
+
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products imaginary parts is below 0.1 %')
```
seongsujeong, line 137:
```suggestion
    print('Check that the percentage of pixels in the difference between reference'
          'and secondary products imaginary parts above the threshold 1.0e-5 is below 0.1 %')
```

seongsujeong (commented):
Thanks for addressing my comments above. My output data have passed the validation with the new script (yay).
My last question is if it is okay to hard-code the thresholds?

And I think it would be helpful if the script displays some information about the comparison (thresholds, ratio of pixels above threshold etc.) when the validation fails. But I don't want to hold this PR too long for that feature, so I would suggest to write an issue and make that happen in the follow-up PRs.

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,32 @@ def compare_cslc_products(file_ref, file_sec):
         return
 
     # Compare amplitude of reference and generated CSLC products
```
seongsujeong, line 112:
```suggestion
    # Compare amplitude of reference and generated CSLC products
    np.seterr(invalid='ignore')
```
In case you want to suppress the warning message like below:
> RuntimeWarning: invalid value encountered in divide

vbrancat:
Thank you for your review @seongsujeong. Regarding of having the threshold exposed as an option, I am a bit conflicted about it. The idea behind the script is that we come up with something that makes the quality of the generated product to be acceptable. Having the threshold to be tunable would allow the users to make that choice (and eventually tune it to make the validation script pass). What would be the advantage of exposing it?

seongsujeong:
> Regarding of having the threshold exposed as an option, I am a bit conflicted about it. The idea behind the script is that we come up with something that makes the quality of the generated product to be acceptable. Having the threshold to be tunable would allow the users to make that choice (and eventually tune it to make the validation script pass). What would be the advantage of exposing it?

Sorry that my question was not so straightforward. I agree that the thresholds / criteria are fixed in the code. What I meant was that we declare some constants in the code  (e.g. `RELATIVE_TOL`, `MAX_RATIO_FAILED`) and use them instead of the `1.0e-5` and `0.001` in the code.

vbrancat:
@LiangJYu do you have time to have a look at this script? @hfattahi if possible, can you make sure that the script does what we agreed for the validation of the CSLC-S1? Merci

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,33 @@ def compare_cslc_products(file_ref, file_sec):
...
+    np.seterr(invalid='ignore')
+    diff_real = np.abs((slc_ref.real - slc_sec.real) / slc_sec.real)
+    diff_imag =  np.abs((slc_ref.imag - slc_sec.imag) / slc_sec.imag)
+
+    # Compute total number of pixels different from nan from ref and sec
+    tot_pixels_ref = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
+    tot_pixels_sec = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
+
+    # Check that total number of pixel is the same
+    assert tot_pixels_ref == tot_pixels_sec
```
LiangJYu, line 122: I don't think comparing number of Nan's is a sufficiently robust check. The following toy example of 2 4x4 rasters (red Nan and green valid) would pass, when I think it ought to fail.
image

I think following per-pixel test would better validate that geolocation is consistent the same between the rasters:
```python
ref_nan = isnan(slc_ref)
sec_nan = isnan(slc_sec)
np.testing.assert_array_equal(ref_nan, sec_nan)
```

I left out `np.abs` because if either real or imaginary component is NaN, `np.nan` evaluates to `True`. e.g.:
```python
In [4]: np.isnan([1, np.nan + np.nan * 1j, np.nan + 1j, 1 + np.nan * 1j])
Out[4]: array([False,  True,  True,  True])
```

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,33 @@ def compare_cslc_products(file_ref, file_sec):
...
     # Compare amplitude of reference and generated CSLC products
-    print('Check mean real part difference between CSLC products is < 1.0e-5')
-    assert np.allclose(slc_ref.real, slc_sec.real,
-                       atol=0.0, rtol=1.0e-5, equal_nan=True)
-    print('Check mean imaginary part difference between CSLC products is < 1.0e-5')
-    assert np.allclose(slc_ref.imag, slc_sec.imag,
-                       atol=0.0, rtol=1.0e-5, equal_nan=True)
+    np.seterr(invalid='ignore')
+    diff_real = np.abs((slc_ref.real - slc_sec.real) / slc_sec.real)
+    diff_imag =  np.abs((slc_ref.imag - slc_sec.imag) / slc_sec.imag)
```
LiangJYu, line 115: `np.abs` redundant here no? The `real` and `imag` properties should not leave any imaginary numbers.

LiangJYu, line 115:
```suggestion
```
Let's move this after NaN pixel comparison. There's no need to perform this computation if the pixel-wise NaN test fails.

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,33 @@ def compare_cslc_products(file_ref, file_sec):
...
+
+    # Compute total number of pixels different from nan from ref and sec
+    tot_pixels_ref = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
+    tot_pixels_sec = np.count_nonzero(~np.isnan(np.abs(slc_ref)))
+
+    # Check that total number of pixel is the same
+    assert tot_pixels_ref == tot_pixels_sec
+
```
LiangJYu, line 123:
```suggestion
    ma_slc_ref = np.ma.masked_array(slc_ref, mask=~ref_nan)
    ma_slc_sec = np.ma.masked_array(slc_sec, mask=~sec_nan)
    diff_real = np.abs((ma_slc_ref .real - slc_ma_slc_sec sec.real) / ma_slc_sec .real)
    diff_imag =  np.abs((ma_slc_ref .imag - ma_slc_sec .imag) / ma_slc_sec .imag)
```
I think it's safer to use masked arrays instead of ignoring all numpy errors. The Nan-mask has already been computed so might as well use it.

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,33 @@ def compare_cslc_products(file_ref, file_sec):
...
+    # Compute the number of pixels in real part above threshold
+    failed_pixels_real = np.count_nonzero(diff_real > 1.0e-5)
+    failed_pixels_imag = np.count_nonzero(diff_imag > 1.0e-5)
```
LiangJYu, line 126:
```suggestion
    max_diff_threshold = 1e-5
    failed_pixels_real = np.count_nonzero(diff_real > max_diff_threshold)
    failed_pixels_imag = np.count_nonzero(diff_imag > max_diff_threshold)
```
Replaced 2 float values with a verbose variable for clarity.

src/compass/utils/validate_cslc.py
```diff
@@ -110,12 +110,33 @@ def compare_cslc_products(file_ref, file_sec):
...
+    percentage_imag = failed_pixels_imag / tot_pixels_ref
+
+    # Check that percentage of pixels above threshold is lower than 0.1 %
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products real parts above the threshold 1.0e-5 is below 0.1 %')
+    assert percentage_real < 0.001
+
+    print('Check that the percentage of pixels in the difference between reference'
+          'and secondary products imaginary parts above the threshold 1.0e-5 is below 0.1 % %')
+    assert percentage_imag < 0.001
```
LiangJYu, line 139:
```suggestion
    percent_fail_threshold = 0.001
    for dtype_op, dtype_str in zip([np.real, np.imag], 
                                   ['real', 'imag']):
        diff = (dtype_op(ma_slc_ref) - dtype_op(ma_slc_secsec)) / dype_op(ma_slc_sec)
        percent_fail = diff / tot_pixels_ref
        # Check that percentage of pixels above threshold is lower than 0.1 %
        print(f'Check that the percentage of pixels in the difference between reference'
              'and secondary products {dtype} parts above the threshold {max_diff_threshold} is below {percent_fail_threshold*100} %')
        assert percent_fail < percent_fail_threshold
```
Percentage failed computation moved into loop to avoid unnecessary repeat computation.

Another approach would be run both real and imag tests and saving the error string should either test fail. This is more thorough since both real and imag pixels are checked. Code as follows:
```python
    percent_fail_threshold = 0.001
    any_percent_fail = False
    err_str = []
    for dtype_op, dtype_str in zip([np.real, np.imag], 
                                   ['real', 'imag']):
        diff = (dtype_op(ma_slc_ref) - dtype_op(ma_slc_sec)) / dype_op(ma_slc_sec)
        percent_fail = diff / tot_pixels_ref
        # Check that percentage of pixels above threshold is lower than 0.1 %
        err_str.append(f'Check that the percentage of pixels in the difference between reference'
              'and secondary products {dtype} parts above the threshold {max_diff_threshold} is below {percent_fail_threshold*100} %')
        any_percent_fail = percent_fail < percent_fail_threshold
    assert any_percent_fail, '\n'.join(err_str)
```

src/compass/utils/validate_cslc.py
vbrancat: uhm, I am a bit confused. My lines of code are comparing pixels that are not NaNs and leaving aside those pixels which are NaNs. With your code, we would compare only NaNs. The idea of the test is that the number of valid pixels (not NaNs) in the reference and secondary RSLCs are the same. Your code would do the opposite, doesn't it?

src/compass/utils/validate_cslc.py
vbrancat: We want to compute the absolute difference between real and/or imaginary parts and we don't care about the sign. In this case, `np.abs` is not redundant. Correct me if I am wrong @seongsujeong

src/compass/utils/validate_cslc.py
LiangJYu: By using `np.count_nonzero`, only the number of non-NaN values is compared and no consideration is given for where the values are located in the array. Both toy rasters have 2 non-zero pixels at different locations and it incorrectly passes the current assert.

`np.testing.assert_array_equal` does an element-by-element comparison of the NaN mask and it would correctly fail with the toy rasters since the non-NaNs are at different places.

src/compass/utils/validate_cslc.py
LiangJYu: Thank you for the explanation. I failed to understand that the absolute difference was being computed.

src/compass/utils/validate_cslc.py
vbrancat: For the last chunk of code, I am not sure what is the intent with 2 items:

1. `percent_fail` is a matrix but we do want a ballpark number describing the percentage of pixels failing to meet the condition, what is the intent here in your code?
2. In the last line with the `assert` condition, `any_percent_fail` is a matrix again and the `assert` condition will fail. Is this intentional?

LiangJYu: The current code will only report an error in the imaginary if the real test pass. i.e.:
```
compute real_percentage
assert real_percentage < percentage_threshold

# check imaginary only if previous assert passes
compute imaginary_percentage
assert imaginary_percentage < percentage_threshold
```

The 2nd chunk of code (when properly coded) will test for both real and imaginary. i.e.:
```
for real_imag in [real, imag]:
    if read_imag_percentage > percentage_threshold:
        save that real or imag failed
        no_fail = False
assert no_fail, error_message
```

Apologies for the errors in my code suggestions. The following should address the failing you listed above:
```python
    # declare thresholds
    pixel_diff_threshold = 1e-5
    percent_fail_threshold = 0.001
    
    # container to capture all possible dtype failures
    # tests passes if empty i.e. no failures were captured
    dtype_fails = []

    # iterate to compute and check %age diff in real and imaginary
    for dtype_op, dtype_str in zip([np.real, np.imag], 
                                   ['real', 'imag']):
        # compute diff per pixel
        diff = (dtype_op(ma_slc_ref) - dtype_op(ma_slc_sec)) / dype_op(ma_slc_sec)
        
        # count number that exceed threshold
        n_diff_gt_threshold = np.count_non_zero(diff > pixel_diff_threshold)
        
        # compute % that fail
        percent_fail = n_diff_gt_threshold / tot_pixels_ref
        
        # Check that percentage of pixels above percent_fail_threshold
        if percent_fail > percent_fail_threshold
            dtype_fails.append(dtype_str)

    # check no fails occurred by check of
    err_str = f'Percentage of pixels in the difference between reference and '
               'secondary products parts above the threshold {pixel_diff_threshold} '
               'is above {percent_fail_threshold*100}% for: '
    assert len(dtype_fails) == 0, err_str + ','.join(dtype_fails)
```

vbrancat:
@LiangJYu would it be fine for you to keep the current implementation? It seems to be clearer albeit more verbose

LiangJYu:
> @LiangJYu would it be fine for you to keep the current implementation? It seems to be clearer albeit more verbose

[This PR](https://github.com/vbrancat/COMPASS/pull/5) fixes the bugs in my suggested code.

vbrancat:
@seongsujeong and @LiangJYu this PR should be good to go. It would be great if you can have another look. Merci

LiangJYu (approved):
LGTM

seongsujeong (approved):
LGTM. Nice work!

PR #85, opened 2023-01-25
Range lut correction

seongsujeong:
This PR addresses the issue regarding the calculation of range LUT, as well as the direction of azimuth LUT for correction.


src/compass/utils/lut.py
```diff
@@ -40,8 +40,8 @@ def cumulative_correction_luts(burst, dem_path,
                                           rg_step=rg_step,
```
scottstanie, line 40: Sorry that  I can't directly comment on the line, but should we remove `scratch_path`? I'm getting an unused variable error from flake8.

scottstanie: alternatively- did you mean to pass `scratch_path` as an argument to `compute_geocoding_correction_luts`?

src/compass/utils/lut.py
```diff
@@ -40,8 +40,8 @@ def cumulative_correction_luts(burst, dem_path,
...
                                           az_step=az_step)
 
-    rg_lut_data = geometrical_steer_doppler.data
-    az_lut_data = bistatic_delay.data + az_fm_mismatch.data
+    rg_lut_data = geometrical_steer_doppler.data * isce3.core.speed_of_light / 2.0
+    az_lut_data = -(bistatic_delay.data + az_fm_mismatch.data)
```
scottstanie, line 44: Maybe we could add some comments on why we're doing the negation for `az_lut_data` and the conversion to meters for `rg_lut_data`?

scottstanie (approved):
The corrections now seem to work again in my test interferogram, so thats great 👍 

My only other comment is that I find the current setup of code somewhat confusing-

From what I understand...
- `compute_geocoding_correction_luts` calls functions defined in `s1reader` to compute different corrections that all return units in seconds. Some of these are for adjusting the azimuth time (which, by the docstring, are the first and third returns, `geometrical_steering_doppler` and `az_fm_mismatch`)
- For the `bistatic_delay` I can't tell what it's supposed to do. In the docstring here it says
> This correction needs to be added to the SLC tagged range time to
        get the corrected range times.
[in s1reader, it says](https://github.com/opera-adt/s1-reader/blob/main/src/s1reader/s1_burst_slc.py#L558-L559)
> This correction needs to be added to the SLC tagged azimuth time
which contradicts it.

Also, the geometrical steer_doppler one says "azimuth" in the docstring,
>     geometrical_steering_doppler: isce3.core.LUT2d:
        LUT2D object of total doppler (geometrical doppler +  steering doppler)
        in seconds as the function of the azimuth time and slant range,
        or range and azimuth indices.
        This correction needs to be added to the SLC tagged azimuth time to
        get the corrected azimuth times.

But we're doing this in the code
```python
rg_lut_data = geometrical_steer_doppler.data * isce3.core.speed_of_light / 2.0
```

So i'm gonna approve this PR now so we can get back to the defaults of COMPASS producing correct outputs, but have some simple way of laying out the corrections we're doing, what they correspond to, and then fix the related documentation on them

LiangJYu:
@seongsujeong I've submitted a [PR](https://github.com/seongsujeong/COMPASS/pull/7/files) to this PR to address @scottstanie comments on documentation.

LiangJYu (approved):
LGTM

vbrancat (approved):

vbrancat:
@seongsujeong please go ahead and merge this PR

PR #87, opened 2023-01-26
Include default values in runconfig written to HDF5 metadata

LiangJYu:
This PR adds default values to runconfig string written to the HDF5 metadata. Currently only the user input runconfig is written to HDF5 metadata.

### DIFF
--- a/src/compass/utils/runconfig.py
+++ b/src/compass/utils/runconfig.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 import os
 from types import SimpleNamespace
 import sys
+import yaml
 
 import isce3
 import journal
@@ -26,6 +27,11 @@ def load_validate_yaml(yaml_path: str, workflow_name: str) -> dict:
         Path to yaml file containing the options to load
     workflow_name: str
         Name of the workflow for which uploading default options
+
+    Returns
+    -------
+    dict
+        Validated user runconfig dict with defaults inserted
     """
     error_channel = journal.error('runconfig.load_validate_yaml')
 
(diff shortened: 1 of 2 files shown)


vbrancat (commented):
LGTM. I do not have any comment on the code. I have not had time to test it. Will do this after finishing some processing on aurora. Will approve after that

src/compass/utils/runconfig.py
```diff
@@ -453,5 +458,5 @@ def to_yaml(self):
         '''Dump runconfig as string to sys.stdout
         '''
         self_as_dict = self.as_dict()
-        yaml = YAML(typ='safe')
-        yaml.dump(self_as_dict, sys.stdout)
+        yaml_obj = YAML(typ='safe')
```
scottstanie, line 461: just checking- any reason for this change?

src/compass/utils/runconfig.py
```diff
@@ -346,15 +352,14 @@ def load_from_yaml(cls, yaml_path: str, workflow_name: str) -> RunConfig:
                 sns.input_file_group.reference_burst.file_path,
                 sns.input_file_group.burst_id)
 
-        # For saving entire file as string to metadata. Stop gap for writing
-        # dict to individual elements to HDF5 metadata
-        with open(yaml_path, 'r') as f_yaml:
-            entire_yaml = f_yaml.read()
+        # For saving entire file with default fill-in as string to metadata.
```
scottstanie, line 355:
```suggestion
        # For saving entire file with defaults filled-in as string to metadata.
```

scottstanie (approved):
I also haven't ran the test yet, but it looks like it should be fine. 
I just wanted to check, since it wasn't obvious to me at the first read
Is the only place where the `cfg` is mutated in the `load_validate_yaml`? 
- if so, maybe we should move the `user_plus_default_yaml_str = yaml.dump(cfg)` right below to indicate that the changing of `cfg` is set
- if not, could we put some comma/indicate in the name somehow that further changes/mutations are done on the cfg? It's just hard to tell from the function where the changes might be happening

src/compass/utils/runconfig.py
LiangJYu: Not necessary, but appended `_obj` to avoid conflict with `import yaml` up top.

PR #90, opened 2023-02-13
s1_geocode_stack.py: set `auto_download=True` when getting orbits

scottstanie:
While talking about some COMPASS testing that @gracebato was doing, we both realized that after installing COMPASS, the first thing we would do is change the geocode stack script to have `auto_download=True` since you always want to get the orbits for a new area.

(This was also related to https://github.com/opera-adt/s1-reader/pull/100, but I thought that fix made more sense to go into s1reader)


vbrancat (approved):
Thanks, @scottstanie I ran into the same problem. Thanks for having fixed that

LiangJYu (approved):
LGTM

PR #93, opened 2023-02-15
remove split spectrum usage from workflows

LiangJYu:
This PR removes split spectrum usage in workflow code and runconfig yaml.

The split spectrum code is still in utils. How do you all feel about removing it?


vbrancat (commented):
It looks good to me codewise. I still did not have a chance to test it but I do not anticipate any problem with it. Will approve after testing.

seongsujeong (approved):
LGTM. Confirmed that the workflow runs as expected on my end.

vbrancat (approved):
Tested and had no issue. Thanks for the quick fix @LiangJYu

Issue #92, opened 2023-02-13
[Bug]: Recompute LUT correction for CSLC product writing


vbrancat:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

In the current pipeline, the LUT corrections are generated twice 
1) to create the cumulative LUTs in slant range and azimuth directions to pass to the `geocodeSlc` core module 2) to generate the LUTs to include inside the generated CSLC-S1 product

This is unnecessary and it might lead to a significant increase in runtime if we decided to have a finer spacing for the LUTs which currently is 200 m x 200 m.

### What did you expect?

I expected the LUT corrections to be computed once.

One option to solve this is to dump the individual corrections to disk in the scratch directory and read them when it is time to allocate them inside the HDF5 product.

vbrancat:
Solved by #94

PR #94, opened 2023-02-17
Avoid compute corrections twice

vbrancat:
This PR proposes an attempt to solve #92 and avoids to compute the correction layers twice.

At the moment, the correction layers are computed 1) to feed them to the geocodeSlc correction module 2) to allocate them inside the CSLC-S1 data product.

In this PR, the corrections are computed once, saved on disk and read when it is time to allocate them in the dataset.
May be there is a smarter way to do this but this PR gives a first stab to the issue.

### DIFF
--- a/src/compass/utils/helpers.py
+++ b/src/compass/utils/helpers.py
@@ -330,12 +330,12 @@ def write_raster(filename, data_list, descriptions,
     Parameters
     ----------
     filename: str
-        Filepath where to store output dataset
-    data_list: list, np.ndarray
+        File path where to store output dataset
+    data_list: list[np.ndarray]
         List of numpy.ndarray to allocate for each
         raster band. All datasets within the list
         are assumed to have the same shape
-    descriptions: list, str
+    descriptions: list[str]
         List of strings containing a description
         for the bands to allocate
     data_type: gdal.dtype
@@ -371,6 +371,3 @@ def write_raster(filename, data_list, descriptions,
 
     out_ds.FlushCache()
     out_ds = None
-
-
-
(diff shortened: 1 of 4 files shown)


vbrancat:
@LiangJYu would you have time to look into this PR? Thanks

src/compass/utils/helpers.py
```diff
@@ -318,3 +318,59 @@ def burst_bboxes_from_db(burst_ids, burst_db_file=None, burst_db_conn=None):
...
+                 data_type=gdal.GDT_Float32, format='ENVI'):
+    '''
+    Write a multiband GDAL-friendly raster to disk.
+    Each dataset allocated in the output file contains
+    a description of the dataset allocated for that band
+
+    Parameters
+    ----------
+    filename: str
+        Filepath where to store output dataset
```
LiangJYu, line 333:
```suggestion
        File path where to store output dataset
```
spelling nit

src/compass/utils/helpers.py
```diff
@@ -318,3 +318,59 @@ def burst_bboxes_from_db(burst_ids, burst_db_file=None, burst_db_conn=None):
...
+    data_list: list, np.ndarray
```
LiangJYu, line 334:
```suggestion
    data_list: list of np.ndarray
```
Change for clarity - `,` separating the two items gives the wrong impression that it's either or.

src/compass/utils/helpers.py
```diff
@@ -318,3 +318,59 @@ def burst_bboxes_from_db(burst_ids, burst_db_file=None, burst_db_conn=None):
...
+        List of numpy.ndarray to allocate for each
+        raster band. All datasets within the list
+        are assumed to have the same shape
+    descriptions: list, str
```
LiangJYu, line 338:
```suggestion
    descriptions: list of str
```
same as above

LiangJYu (approved):
LGTM - only formatting and documentation comments

src/compass/utils/helpers.py
LiangJYu:
```suggestion
    data_list: list[np.ndarray]
```
This also works and I think is convention we've been using in COMPASS.

vbrancat:
@seongsujeong Would you mind also having a look at this PR?

seongsujeong (approved):
LGTM

PR #97, opened 2023-02-24
geocode stack adjustments for a gzipped-json burst db

scottstanie:
This PR has a few changes that came up while testing out the COMPASS wrapper workflow https://github.com/opera-adt/sweets

The main changes are to allow a format of the burst db that's a simple JSON file instead of a sqlite database. Since it takes <1sec to load 1M rows into memory, there's little noticeable difference in the s1_geocode_stack runtime. The nice part is the DB is only ~15 MB when it's a gzipped JSON file.

Something that should be sorted out before this is merged is the burst_db repo, where this was implemented https://github.com/opera-adt/burst_db/compare/main...scottstanie:burst_db:test-json-gz
But really we can store the v1 version of the DB somewhere easy to download because it's so small.

- Also allows us to point to either a list of pre-unzipped .SAFE directories, as well as .zip files.

### DIFF
--- a/src/compass/s1_geocode_stack.py
+++ b/src/compass/s1_geocode_stack.py
@@ -77,7 +77,8 @@ def create_parser():
 
 
 def generate_burst_map(zip_files, orbit_dir, output_epsg=None, bbox=None,
-                       bbox_epsg=4326, burst_db_file=DEFAULT_BURST_DB_FILE):
+                       bbox_epsg=4326, burst_db_file=DEFAULT_BURST_DB_FILE,
+                       min_bbox_intersect=0.01):
     """Generates a dataframe of geogrid infos for each burst ID in `zip_files`.
 
     Parameters
@@ -116,7 +117,7 @@ def generate_burst_map(zip_files, orbit_dir, output_epsg=None, bbox=None,
             ref_bursts = load_bursts(zip_file, orbit_path, subswath)
             for burst in ref_bursts:
                 epsg, bbox_utm = _get_burst_epsg_and_bbox(
-                    burst, output_epsg, bbox, bbox_epsg, burst_db_file
+                    burst, output_epsg, bbox, bbox_epsg, burst_db_file, min_intersect=min_bbox_intersect
                 )
                 if epsg is None:  # Flag for skipping burst
                     continue
@@ -135,7 +136,7 @@ def generate_burst_map(zip_files, orbit_dir, output_epsg=None, bbox=None,
 
 
 def _get_burst_epsg_and_bbox(burst, output_epsg, bbox, bbox_epsg, burst_db_file,
-                             min_intersect=0.05):
+                             min_intersect=0.01):
     """Returns the EPSG code and bounding box for a burst.
 
     Uses specified `bbox` if provided; otherwise, uses burst database (if available).
@@ -163,7 +164,8 @@ def _get_burst_epsg_and_bbox(burst, output_epsg, bbox, bbox_epsg, burst_db_file,
         bbox_utm_geom = geometry.box(*bbox_utm)
         burst_intersection = burst_border_utm.intersection(bbox_utm_geom)
         # Skip this burst if it doesn't overlap enough with specified bbox
-        intersect_pct = burst_intersection.area / burst_border_utm.area
+        total_area = min(bbox_utm_geom.area, burst_border_utm.area)
+        intersect_pct = burst_intersection.area / total_area
         if intersect_pct < min_intersect:
             return None, None
     else:
--- a/src/compass/s1_geocode_stack.py
+++ b/src/compass/s1_geocode_stack.py
@@ -70,6 +70,9 @@ def create_parser():
     optional.add_argument('-m', '--metadata', action='store_true',
                           help='If flat is set, generates radar metadata layers for each'
                                ' burst stack (see rdr2geo processing options)')
+    optional.add_argument('--unzipped', action='store_true',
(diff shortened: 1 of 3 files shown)


src/compass/s1_geocode_stack.py
```diff
@@ -70,6 +70,9 @@ def create_parser():
     optional.add_argument('-m', '--metadata', action='store_true',
                           help='If flat is set, generates radar metadata layers for each'
                                ' burst stack (see rdr2geo processing options)')
+    optional.add_argument('--unzipped', action='store_true',
```
vbrancat, line 73: I am a bit confused here (I might need a bit more context). I have never unzipped my SAFE file when using COMPASS and specifically stack processor. What is the aim of this option?

src/compass/utils/helpers.py
```diff
@@ -219,19 +223,27 @@ def _convert_to_utm(points_xy, epsg_src, epsg_dst):
     return list(zip(xt, yt))
 
 
-def burst_bbox_from_db(burst_id, burst_db_file=None, burst_db_conn=None):
+@functools.lru_cache(maxsize=2)
+def _read_burst_json(burst_db_file: Union[str, Path]) -> dict:
```
vbrancat, line 227: If I remember correctly, @seongsujeong performed an analysis related to how much time we need to load into memory a burst data base allocated in SQlite or JSON. If I remember the conclusion correctly, we decided to go for Sqlite because of runtime performance. Can you cack up this choice with some numbers?

vbrancat (commented):
@scottstanie let's discuss about changing the format of the burst database to JSON. I remember that @seongsujeong did some analysis on what was the best format to use for the burst database and we landed on the SQLite because of performance. Also, The trackframe database should have its on repository and it should be versioned. This would allow us to cut a release every time we need to make a delivery. What do you think?

src/compass/s1_geocode_stack.py
scottstanie: The aim is for quicker runtime when geocoding many bursts from one granule. Here's the numbers I got when I checked geocoding one burst using the .zip file vs the unzipped .SAFE folder.

| Threads | Zipped runtime (s) | Unzipped runtime (s) |
| ------- | ------------------ | -------------------- |
| 8       | 50                 | 39                   |
| 16      | 34                 | 22                   |
| 32      | 25                 | 14                   |
| 48      | 21                 | 10                   |


The `Threads` were set by doing, e.g., `$ OMP_NUM_THREADS=16 s1_cslc.py test-zipped.yaml`

Geocoding all the bursts in a stack of 20 dates means that if you can us 48 threads, it'll take about `21 * 30 * 30 = 210` minutes using the .zip, and 100 minutes using the .SAFE.  
The 20 dates took 3.5 - 4 minutes to unzip the VV polarization images/metadata.

scottstanie:
Also, if it's helpful to separate discussion between adding the ability to run on .SAFE files and using a .json.gz format, I can close this PR and break them out.

src/compass/s1_geocode_stack.py
vbrancat: Thanks for the explanation. Does the unzipped runtime include the time to unzip the SAFE file itself? I would expect that the unzipped runtime just includes the time to run the CSLC-S1 workflow on an unzipped SAFE file.

src/compass/utils/helpers.py
vbrancat: I see. So summarizing what you are saying, we can provide the burst database (BDB) as a zipped JSON file. If we do so, we would have a flat 1-3 s to unzip the actual file and once the dictionary is loaded in memory, the other calls should be faster. Correct?

When  I was developing the stack processor, I remember experiencing some delays in scanning very big dictionaries. If I remember correctly, this was one of the analyses that also @seongsujeong (maybe, you can share your presentation in this PR, so that our memory will be refreshed).

I would not be too worried in including the time to download the BDB. For production 1) we do not use the stack processor 2) we assumed that the BDB is always staged somewhere and ready to use.

At the moment, the only implication I see in changing the file format of the BDB is how fast we are in scanning the all BDB for very large stacks (>100 acquisitions). Once we are sure, we are not introducing significant lags, I am fine with it

vbrancat:
> 

Breaking those discussions in two separate PRs would be extremely helpful (more work on you though, :( )

src/compass/s1_geocode_stack.py
scottstanie: Sorry i smushed that part together up there:

The 20 dates took 3.5 - 4 minutes to unzip unto SAFE files. I selectively unzipped the VV polarization images/metadata only, not the VH.

src/compass/utils/helpers.py
scottstanie: I think seongsu's timings might have been checking what was the time to load *and* search for one item? We can check when he posts it.

The lookup in a dict, once it's loaded, should be in the microsecond range (since it's an O(1) operation) no matter how big the dict is.

Here's the lookup time for a 10M dict:
```python console
In [1]: import numpy as np
In [2]: large_dict = {k: k+1 for k in range(10_000_000)}
In [3]: idxs = (1e7 * np.random.rand(200)).astype(int)  # 200 random idxs from 0 to 1e7

In [4]: %timeit [large_dict[k] for k in idxs]
29 µs ± 932 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

scottstanie:
Closing in favor of separate incoming PRs

PR #99, opened 2023-03-02
Geocode stack additions

scottstanie:
broken off from the previous PR with the burst DB changes are

1. the ability to point to unzipped SAFE directories
2. exposing turning off the correction LUTs for the geocode stack script.

Also minor fixes to docstrings/default args, and using keyword args on the function calls which had gotten very long (I didn't want to mess up where i was inserting the new args).

The correction addition was prompted by the first test of `sweets` that @hfattahi did. He saw a burst discontinuity for one burst in a frame:

image

I ran it with a different processor (the stanford geocoding one) that applies no corrections, and it looks like this in that area:
image

The strip of blurry decorrelation is because that processor doesn't account for shifting burst footprint, but notice that there phase on either side looks continuous, while the COMPASS one looks completely different.

### DIFF
--- a/src/compass/s1_geocode_stack.py
+++ b/src/compass/s1_geocode_stack.py
@@ -459,13 +459,13 @@ def run(slc_dir, dem_file, burst_id=None, common_bursts_only=False, start_date=N
         do_metadata = enable_metadata and (row.Index in first_rows)
         runconfig_path = create_runconfig(
             row,
-            dem_file,
-            work_dir,
-            flatten,
-            pol,
-            x_spac,
-            y_spac,
-            do_metadata,
+            dem_file=dem_file,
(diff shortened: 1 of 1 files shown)


hfattahi:
Thanks @scottstanie. I assume the discontinuity was at two frame boundaries. Were you able to confirm? 
For completeness would you please post the intreferogram without lut corrections.

hfattahi (approved):
LGTM! Thanks Scott.

LiangJYu (approved):
LGTM

scottstanie:
@hfattahi  no I don't think it's a frame boundary, see here for plots https://github.com/opera-adt/COMPASS/issues/100#issuecomment-1452437338

PR #91, opened 2023-02-13
Solid Earth Tides correction

vbrancat:
This PR adds the Solid Earth Tides (SET) to the list of Look UP Table (LUT) corrections given to the `geocodeSlc` cor module. SET are computed using `pySolid`. The generated LUT in radar coordinates along slant range and azimuth are also added to the CSLC product.

Note that `pySolid` is not added as a dependency as the next `isce3` conda release should already support this dependency.

PR #102, opened 2023-03-07
Ionosphere correction

seongsujeong:
This PR adds ionosphere correction feature to COMPASS. Currently I label this PR `work in progress` because there are some points that needs to be coordinated with the pending PRs in COMPASS.
Below are few notes:

1. For comparison purpose (with the ISCE3 equivalent in PR #1251 in the internal GitHub), static troposphere correction was added. It is TBD whether or not to keep it.
2. IONEX file is provided as one of dynamic ancillary files in runconfig.
3. Current behavior regarding IONEX is that it raises an error when `correction_luts` is enabled but IONEX file was not provided, or does not exist.
4. It re-uses lat / lon / incidence angle rasters in radar grid. The current commit in this PR runs `rdr2geo` when calculating azimuth FM rate mismatch.
    4.1. Eventually solid earth tide computation (in the PR #91 ) will run `rdr2geo`. This run will generate `latitude`, `longitude`, `height`, `incidence angle`, and `heading` to be used for other corrections, which are:
    - Solid Earth Tide
    - azimuth FM mismatch rate
    - ionosphere correction
    - troposphere correction in python level (Either static or weather model-based)


seongsujeong:
@vbrancat @LiangJYu @hfattahi I think the PR is ready for review.

src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -17,6 +17,8 @@ runconfig:
       dynamic_ancillary_file_group:
           # Digital elevation model
           dem_file:
+          # IONEX file for ionosphere correction
+          ionex_file:
```
vbrancat, line 21: Ionex is a file format, let's rename this to something more descriptive e.g. `tec_file `. Also let's add documentation of what this file is e.g. GNSS total electron content file in Ionex format

src/compass/utils/lut.py
```diff
@@ -76,16 +77,16 @@ def cumulative_correction_luts(burst, dem_path,
     output_path = f'{scratch_path}/corrections'
     os.makedirs(output_path, exist_ok=True)
     data_list = [geometry_doppler, bistatic_delay.data, az_fm_mismatch.data,
-                 tide_rg]
+                 tide_rg, ionosphere]
     descr = ['geometrical doppler', 'bistatic delay', 'azimuth FM rate mismatch',
-             'slant range Solid Earth tides']
+             'slant range Solid Earth tides', 'ionospheric delay']
```
vbrancat, line 82: Let's specify that the ionosphere is along the LOS.

src/compass/utils/lut.py
```diff
@@ -362,3 +367,55 @@ def resample_set(geo_tide, pts_src, pts_dest):
                    bounds_error=False, fill_value=0)
     rdr_tide = rgi_func(pts_dest)
     return rdr_tide
+
+
+def ionosphere_delay(sensing_time, wavelength,
```
vbrancat, line 372: The input of the function should be generic. I assume here you need a date to be provided in a certain format not necessarily the burst sensing start.

src/compass/utils/lut.py
```diff
@@ -362,3 +367,55 @@ def resample_set(geo_tide, pts_src, pts_dest):
...
+    lon_arr: numpy.ndarray
+        array of longitude in radar grid
+    lat_arr: numpy.ndarray
+        array of latitude in radar grid
+    inc_arr: numpy.ndarray
+        array of incidence angle in radar grid
+
+    Returns
+    -------
+    slant_range_delay: np.ndarray
```
vbrancat, line 392: Let's rename this variable to reflect that this is a LOS ionosphere delay

src/compass/utils/lut.py
```diff
@@ -362,3 +367,55 @@ def resample_set(geo_tide, pts_src, pts_dest):
...
+        Ionospheric delay in slant range
+    '''
+
+    if not ionex_path:
```
vbrancat, line 396: It is ok to have this check inside the function but we need to discuss what is the default behavior. If a user wants to generate a CSLC-S1 does he/she need to always provide a TEC file?

src/compass/utils/lut.py
```diff
@@ -362,3 +367,55 @@ def resample_set(geo_tide, pts_src, pts_dest):
...
+        raise RuntimeError('LUT correction was enabled, '
```
vbrancat, line 397: Irrespective of the setting for the corrections, if `ionex_path` is not provided, we cannot perform the correction. The error message should reflect that and be independent of the correction settings. Also, let's check directly that the path exists

src/compass/utils/lut.py
```diff
@@ -362,3 +367,55 @@ def resample_set(geo_tide, pts_src, pts_dest):
...
+
+    ionex_val = get_ionex_value(ionex_path,
+                                utc_tod_sec,
+                                lat_arr.flatten(),
+                                lon_arr.flatten())
+
+    ionex_val = ionex_val.reshape(lon_arr.shape)
+
+    freq_sensor = isce3.core.speed_of_light / wavelength
+    electron_per_sqm = ionex_val * 1e16
```
vbrancat, line 415: Let's add a comment why we need to do this conversion

src/compass/utils/h5_helpers.py
```diff
@@ -608,6 +608,9 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
             Meta('los_solid_earth_tides', ds.GetRasterBand(4).ReadAsArray(),
                  f'Solid Earth tides (range) {desc}',
                  {'units': 'meters'}),
+            Meta('ionospheric_delay', ds.GetRasterBand(5).ReadAsArray(),
+                 f'Ionospheric delay {desc}',
```
vbrancat, line 612: Let's add the direction of the ionosphere delay

src/compass/utils/lut.py
```diff
@@ -362,3 +367,55 @@ def resample_set(geo_tide, pts_src, pts_dest):
...
+        array of longitude in radar grid
+    lat_arr: numpy.ndarray
+        array of latitude in radar grid
+    inc_arr: numpy.ndarray
+        array of incidence angle in radar grid
+
+    Returns
+    -------
+    slant_range_delay: np.ndarray
+        Ionospheric delay in slant range
```
vbrancat, line 393: Let's add the unit for this delay

src/compass/utils/lut.py
```diff
@@ -362,3 +367,55 @@ def resample_set(geo_tide, pts_src, pts_dest):
...
+    ----------
+    time_sensing: datetime.datetime
+        Sensing time of burst
+    wavelength: float
+        Wavelength of the signal
+    lon_arr: numpy.ndarray
+        array of longitude in radar grid
+    lat_arr: numpy.ndarray
+        array of latitude in radar grid
+    inc_arr: numpy.ndarray
```
vbrancat, line 387: Let's add unit for all the input variables of this function

vbrancat (commented):
Thank you @seongsujeong. I did not have time to test the PR, unfortunately. However, I provided some comments. Do you mind sharing the TEC file for Rosamond. This might make testing much faster. Thank you :)

src/compass/utils/lut.py
LiangJYu: [`isce3`](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/4694efe362713b3af1d374270ebfaebf469c966a/python/packages/isce3/atmosphere/tec_product.py#L15) uses UTC time.

Also, I think this function is a better fit in [iono.py](https://github.com/opera-adt/COMPASS/blob/main/src/compass/utils/iono.py).

LiangJYu (commented):
Not much to add beyond what @vbrancat already commented on.

src/compass/utils/lut.py
seongsujeong: Good idea. For consistency, what do you think about adding `azimuth` or `slant range` for the first three corrections? i.e. `geometrical doppler`, `bistatic delay`, and `azimuth FM rate mismatch`

src/compass/utils/lut.py
seongsujeong: Thanks for the suggestion. I agree the placement of this function, and generic input for the time component of this function. For that I've moved the function into `iono.py`, and renamed `sensing_time` into `utc_time`
For the data type of `utc_time` I think I prefer to keep `datetime.datetime`. What do you think?

src/compass/utils/lut.py
seongsujeong: I agree that the error message should be independent to the correction setting. Please take a look how the revised message looks like.

For file existence check, I think we might better to keep the current pattern because `os.path.exists(None)` will raise an error like below. Also this early-return pattern will make it easier to read.

> os.path.exists(None)
  File "<stdin>", line 1, in <module>
  File "/Users/jeong/miniconda3/lib/python3.9/genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType

src/compass/utils/lut.py
vbrancat: I agree, it is a good idea.

src/compass/utils/lut.py
vbrancat: Yeah, I like having the data type of `utc_time` to be `datatime.datetime` (let's clarify it in a comment or in the docstring), what do you think @LiangJYu ?

src/compass/utils/lut.py
vbrancat: Let's resume the discussion on the default behavior in our next meeting. In the meantime, we do not want the user to get until here if the TEC file has not been provided. I recommend adding a check in `geo_runconfig.py` similarly to what has been done here https://github.com/vbrancat/COMPASS/blob/016eb4d762b440ba0609e9a8f5fd495a14e8f55a/src/compass/utils/geo_runconfig.py#L73 so that we stop the SAS immediately if the TEC file is not provided

src/compass/utils/lut.py
LiangJYu: `datetime.datetime` sounds good to me.

LiangJYu:
Not immediately relevant to this PR: @seongsujeong @vbrancat what do you all think [consolidating ionosphere functionality](https://github.com/opera-adt/COMPASS/discussions/104) (and maybe other atmospheric corrections) within ISCE3?

src/compass/utils/lut.py
vbrancat: With respect to the SAS default behavior, just a heads up, I remember we decided to support an interface where the provided path to the TEC file might be `None`

src/compass/utils/lut.py
```diff
@@ -27,6 +27,8 @@ def cumulative_correction_luts(burst, dem_path,
         Sentinel-1 A/B burst SLC object
     dem_path: str
         Path to the DEM file
+    tec_path: str
+        Path to the TEC file for ionosphere correction
```
vbrancat, line 31: Let's specify the file format for the expected TEC file

src/compass/utils/lut.py
```diff
@@ -44,16 +46,17 @@ def cumulative_correction_luts(burst, dem_path,
         and slant range
     '''
     # Get individual LUTs
-    geometrical_steer_doppler, bistatic_delay, az_fm_mismatch, [tide_rg, _]= \
+    geometrical_steer_doppler, bistatic_delay, az_fm_mismatch, [tide_rg, _], ionosphere = \
```
vbrancat, line 49: let's identify the variable as `los_ionosphere`

src/compass/utils/lut.py
```diff
@@ -97,9 +100,11 @@ def compute_geocoding_correction_luts(burst, dem_path,
         S1-A/B burst object
     dem_path: str
         Path to the DEM required for azimuth FM rate mismatch.
-    xstep: int
+    tec_path: str
+        Path to the TEC file for ionosphere correction
+    rg_step: int
         LUT spacing along x/slant range in meters
-    ystep: int
+    az_ystep: float
```
vbrancat, line 107: This does not reflect the name of the function variable

vbrancat (commented):
Some more small comments

src/compass/utils/lut.py
LiangJYu:
```suggestion
    az_step: float
```

LiangJYu (approved):
LGTM nothing to add beyond what @vbrancat commented on.

src/compass/utils/lut.py
seongsujeong: @vbrancat  Thanks for the suggestion and information. I've added the check in `geo_runconfig.py` for TEC file. Also, in `iono.ionosphere_delay()`, I've modified it to return a zero-filled array when `tec_file` is `None`

```python
if not tec_file:
    print('"tec_file" was not provided. '
             'Ionosphere correction will not be applied.')
    return np.zeros(lon_arr.shape)
```

src/compass/utils/lut.py
seongsujeong: Addressed as mentioned in the comment above.

src/compass/utils/lut.py
seongsujeong: Thanks for catching this. That was a typo.

vbrancat (approved):
LGTM, thanks @seongsujeong

PR #103, opened 2023-03-08
Weather model-based troposphere delay LUTs

vbrancat:
This PR adds the computation of the model-based troposphere delay computation to the generation of a CSLC-S1 product.

The wet and dry troposphere delays are computed using the RAiDER software. Note:

1. RAiDER supports only preprocessed troposphere weather model files. You will need to download the desired weather model using RAiDER. A script can be provided to exercise this functionality.
2. The RAiDER LOS object to compute the LOS troposphere delay requires the orbit file. To bypass that, the PR computes the Zenith delay and projects it on to the LOS.


src/compass/utils/h5_helpers.py
```diff
@@ -609,6 +618,27 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
...
+            elif delay_type == 'wet':
+                correction_items.append(Meta('wet_los_troposphere_delay',
+                                             ds.GetRasterBand(5).ReadAsArray(),
+                                             f'Wet LOS troposphere delay {desc}',
+                                             {'units': 'meters'}))
+            else:
+                correction_items.append(Meta('dry_los_troposphere_delay',
+                                             ds.GetRasterBand(6).ReadAsArray(),
+                                             f'Dry LOS troposphere delay {desc}',
+                                             {'units': 'meters'}))
```
LiangJYu, line 640:
```suggestion
        if weather_model_path is not None:
            if 'wet' in delay_type:
                correction_items.append(Meta('wet_los_troposphere_delay',
                                             ds.GetRasterBand(5).ReadAsArray(),
                                             f'Wet LOS troposphere delay {desc}',
                                             {'units': 'meters'}))
            if 'dry' in delay_type:
                correction_items.append(Meta('dry_los_troposphere_delay',
                                             ds.GetRasterBand(6).ReadAsArray(),
                                             f'Dry LOS troposphere delay {desc}',
                                             {'units': 'meters'}))
```
Less repetition

src/compass/utils/lut.py
```diff
@@ -44,17 +57,28 @@ def cumulative_correction_luts(burst, dem_path,
...
     geometry_doppler = geometrical_steer_doppler.data * isce3.core.speed_of_light * 0.5
     rg_lut_data = geometry_doppler + tide_rg
 
+    # Add troposphere delay to range LUT
+    if delay_type == 'wet_dry':
+        rg_lut_data += wet_los_tropo + dry_los_tropo
+    elif delay_type == 'wet':
+        rg_lut_data += wet_los_tropo
+    else:
+        rg_lut_data += dry_los_tropo
```
LiangJYu, line 80:
```suggestion
    if 'wet' in delay_type:
        rg_lut_data += wet_los_tropo
    if 'dry' in delay_type:
        rg_lut_data += dry_los_tropo
```
Less repitition

src/compass/utils/lut.py
```diff
@@ -80,32 +104,49 @@ def cumulative_correction_luts(burst, dem_path,
...
 
+    if delay_type == 'wet_dry':
+        data_list.extend([wet_los_tropo, dry_los_tropo])
+        descr.extend(['wet LOS troposphere', 'dry_los_troposphere'])
+    elif delay_type == 'wet':
+        data_list.append(wet_los_tropo)
+        descr.append('wet LOS troposphere')
+    else:
+        data_list.append(dry_los_tropo)
+        descr.append('dry LOS troposphere')
```
LiangJYu, line 115:
```suggestion
    if delay_type == 'wet':
        data_list.append(wet_los_tropo)
        descr.append('wet LOS troposphere')
    if delay_type == 'dry':
        data_list.append(dry_los_tropo)
        descr.append('dry LOS troposphere')
```
Less repetition

src/compass/utils/lut.py
```diff
@@ -131,7 +172,13 @@ def compute_geocoding_correction_luts(burst, dem_path,
         List of numpy.ndarray containing SET in slant range and azimuth directions
         in meters. These corrections need to be added to the slC tagged azimuth
         and slant range times.
+
+    [wet_los_tropo, dry_los_tropo]: list, np.ndarray
```
LiangJYu, line 176:
```suggestion
    [wet_los_tropo, dry_los_tropo]: list[np.ndarray]
```
This is less ambiguous. Also update SET corrections above the same way.

src/compass/utils/lut.py
```diff
@@ -183,8 +230,30 @@ def compute_geocoding_correction_luts(burst, dem_path,
     rg_set = resize(rg_set_temp, out_shape, **kwargs)
     az_set = resize(az_set_temp, out_shape, **kwargs)
 
+    # Compute wet and dry troposphere delays using RAiDER
+    wet_los_tropo, dry_los_tropo = np.zeros(out_shape), np.zeros(out_shape)
```
LiangJYu, line 234:
```suggestion
    wet_los_tropo = np.zeros(out_shape)
    dry_los_tropo = np.zeros(out_shape)
```
Nit - one line per assignment

Or with less repetition and single line
```suggestion
    wet_los_tropo, dry_los_tropo = [np.zeros(out_shape) for _ in range(2)]
```

LiangJYu (commented):
Update current docker/specifile.txt with the attached one.
[specifile.txt](https://github.com/opera-adt/COMPASS/files/10936791/specifile.txt)

vbrancat:
@LiangJYu not sure why the CI is failing. It mentions a bug in conda.

LiangJYu (approved):
LGTM

src/compass/utils/lut.py
```diff
@@ -183,8 +225,30 @@ def compute_geocoding_correction_luts(burst, dem_path,
...
+
+        # Compute the troposphere delay along the Zenith
+        zen_wet, zen_dry = tropo_delay(burst.sensing_start,
+                                       weather_model_path,
+                                       aoi, los)
+
+        # RaiDER delay is one-way only. Get the LOS delay my multiplying
+        # by the incidence angle
+        wet_los_tropo = 2.0 * zen_wet * np.cos(np.deg2rad(inc_angle))
+        dry_los_tropo = 2.0 * zen_dry * np.cos(np.deg2rad(inc_angle))
```
scottstanie, line 248: `zenith = LOS * cos(incidence) -> LOS = zenith / cos`

```suggestion
        wet_los_tropo = 2.0 * zen_wet / np.cos(np.deg2rad(inc_angle))
        dry_los_tropo = 2.0 * zen_dry / np.cos(np.deg2rad(inc_angle))
```

docker/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN ${CONDA_PREFIX}/bin/conda init bash
 COPY --chown=compass_user:compass_user . /home/compass_user/OPERA/COMPASS
 
 # create CONDA environment
-RUN conda create --name "COMPASS" --file /home/compass_user/OPERA/COMPASS/docker/specifile.txt
+RUN conda create --name "COMPASS" --file /home/compass_user/OPERA/COMPASS/docker/specifile.txt -c conda-forge --override-channels
```
seongsujeong, line 33: A question: It seems all dependencies in `specifile.txt` has `conda-forge` in their URLs (correct me if I missed something). Are those options are really necessary, or just to make sure that all packages are from conda forge?

src/compass/utils/h5_helpers.py
```diff
@@ -609,6 +618,18 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
...
                  {'units': 'meters'}),
         ]
+        if weather_model_path is not None:
+            if 'wet' in delay_type:
+                correction_items.append(Meta('wet_los_troposphere_delay',
+                                             ds.GetRasterBand(5).ReadAsArray(),
+                                             f'Wet LOS troposphere delay {desc}',
+                                             {'units': 'meters'}))
+            if 'dry' in delay_type:
+                correction_items.append(Meta('dry_los_troposphere_delay',
```
seongsujeong, line 628: Does not need to be addressed in this PR, but reminding that we will need to rearrange the order of the corrections to write into the output HDF5 file because we have SET azimuth, and ionosphere not in place yet.
I'll submit a WIP PR to azimuth SET conversion, and let's utilize that PR to finalize all the corrections we have implemented so far.

seongsujeong (approved):
Looks good to me. Just a question and a reminder.

docker/Dockerfile
vbrancat: Deferring this question to @LiangJYu who provided the `specifile.txt`

LiangJYu: Good catch @seongsujeong. You're correct what I added here is unnecessary. I got the specifile.txt confused with requirements.txt.

vbrancat: @LiangJYu will you take care of this?

src/compass/utils/h5_helpers.py
vbrancat: May be we can open an issue to track this down @seongsujeong ?

scottstanie (approved):

src/compass/utils/h5_helpers.py
seongsujeong: The tasks for the finalization will be tracked here: 
#106

docker/Dockerfile
LiangJYu, line 33:
```suggestion
RUN conda create --name "COMPASS" --file /home/compass_user/OPERA/COMPASS/docker/specifile.txt
```
Remove unnecessary flags that forced conda forge channel

src/compass/utils/geo_runconfig.py
```diff
@@ -64,6 +65,14 @@ def load_from_yaml(cls, yaml_path: str, workflow_name: str) -> GeoRunConfig:
         geocoding_dict = groups_cfg['processing']['geocoding']
         check_geocode_dict(geocoding_dict)
 
+        # Check troposphere weather model file if not None. This
+        # troposphere correction is applied only if this file is not None
+        weather_model_path = groups_cfg['dynamic_ancillary_file_group'][
+            'weather_model_file'
+        ]
+        if weather_model_path is not None:
+            check_file_path(weather_model_path)
```
seongsujeong, line 74: In case the weather model file is not provided, we can still apply troposphere delay using static model in python level because we have all ingredients for the calculation. What do you think?

vbrancat: I agree. I think we should apply the static troposphere correction regardless. Only, have we decided if we want to proceed with the computation of the static delay at the core module or at a Python level? I am in favor of the second, it should not be that expensive to compute anyway

PR #107, opened 2023-03-14
Correct inconsistencies inside CSLC-S1 metadata

vbrancat:
This PR tries to correct some inconsistencies in the allocation of the product metadata. Also, it tries to improve the description of some allocated metadata fields.


LiangJYu (approved):
LGTM

src/compass/utils/h5_helpers.py
```diff
@@ -370,10 +370,10 @@ def identity_to_h5group(dst_group, burst):
             'Azimuth stop time of product'),
         Meta('list_of_frequencies', ['A'],
              'List of frequency layers available in the product'),  # T)C
-        Meta('is_geocoded', True, 'Flag to indicate radar geometry or geocoded product'),
-        Meta('is_urgent_observation', False,
+        Meta('is_geocoded', 'True', 'Flag to indicate radar geometry or geocoded product'),
+        Meta('is_urgent_observation', 'False',
              'List of booleans indicating if datatakes are nominal or urgent'),
```
scottstanie, line 375: Is this a "List of booleans"?
Would we need to change this to `string of "True" or "False" if we're making it a string?

src/compass/utils/h5_helpers.py
```diff
@@ -370,10 +370,10 @@ def identity_to_h5group(dst_group, burst):
             'Azimuth stop time of product'),
         Meta('list_of_frequencies', ['A'],
              'List of frequency layers available in the product'),  # T)C
-        Meta('is_geocoded', True, 'Flag to indicate radar geometry or geocoded product'),
-        Meta('is_urgent_observation', False,
+        Meta('is_geocoded', 'True', 'Flag to indicate radar geometry or geocoded product'),
```
scottstanie, line 373: Same with this, I would assume that "Flag" is a boolean?

scottstanie (approved):
Looks like good clarifications, but just checking- why are we converting the boolean `True/False`s to strings?

src/compass/utils/h5_helpers.py
vbrancat: I suspect that the "list" comes from NISAR but it is not really needed for CSLC-S1 as each product will contain only a single burst.

scottstanie:
> @scottstanie I have received a comment from some product reviewers that a string is more intuitive than a flag. Personally, I find the flag more intuitive. Since you are closer to the user community than I am, what would you recommend?

If others were confused by it being boolean (maybe it got read as an integer for them or something?) then I don't have any real strong objections. Neither field looks particularly important for usual processing, they're more informational

PR #109, opened 2023-03-14
Populate CSLC-S1 product_version

vbrancat:
This PR populates the metadata identification field `product_version`. This field is populated by grabbing the corresponding value in the CSLC-S1 runconfiguration file.


LiangJYu (approved):
LGTM

src/compass/utils/geo_runconfig.py
```diff
@@ -92,6 +92,10 @@ def load_from_yaml(cls, yaml_path: str, workflow_name: str) -> GeoRunConfig:
         return cls(cfg['runconfig']['name'], sns, bursts, empty_ref_dict,
                    user_plus_default_yaml_str, output_paths, geogrids)
 
+    @property
+    def product_path(self) -> dict:
+        return self.groups.product_path_group
+
```
scottstanie, line 98: This seems confusing to call it `product_path` in the `geo_runconfig`, but the same name in the `runconfig` is something else:

https://github.com/opera-adt/COMPASS/blob/cda14a3a47c1f7afb00a6cb34455043c9b33a79a/src/compass/utils/runconfig.py#L387-L389

vbrancat: Let me rename it as `product_group`, it might be more appropriate

scottstanie (approved):

PR #112, opened 2023-03-15
Fix s1_geocode_metadata crash

LiangJYu:
This PR fixes #110 
- fix s1_geocode_metadata incorrect usage of block params
- enable s1_geocode_metadata in unit tests

PR #113, opened 2023-03-17
Apply static troposphere model when weather model file is not provided

seongsujeong:
This PR is to add a feature that applies static troposphere when weather model was not provided in runconfig. In case the weather model file is provided, the tropospheric correction is based on the weather model.
The static troposphere LUT can be calculated from incidence angle and surface height, which are calculated in `lut.py`.

### DIFF
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -614,16 +614,19 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
         Meta('los_solid_earth_tides', ds.GetRasterBand(4).ReadAsArray(),
              f'Solid Earth tides (range) {desc}',
              {'units': 'meters'}),
+        Meta('los_ionospheric_delay', ds.GetRasterBand(5).ReadAsArray(),
+             f'Ionospheric delay (range) {desc}',
+             {'units': 'meters'}),
     ]
     if weather_model_path is not None:
         if 'wet' in delay_type:
             correction_items.append(Meta('wet_los_troposphere_delay',
-                                         ds.GetRasterBand(5).ReadAsArray(),
+                                         ds.GetRasterBand(6).ReadAsArray(),
                                          f'Wet LOS troposphere delay {desc}',
                                          {'units': 'meters'}))
         if 'dry' in delay_type:
             correction_items.append(Meta('dry_los_troposphere_delay',
-                                         ds.GetRasterBand(6).ReadAsArray(),
+                                         ds.GetRasterBand(7).ReadAsArray(),
                                          f'Dry LOS troposphere delay {desc}',
                                          {'units': 'meters'}))
 
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -617,7 +617,13 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
                  f'Ionospheric delay (range) {desc}',
                  {'units': 'meters'}),
         ]
-        if weather_model_path is not None:
+
+        if weather_model_path is None:
+            correction_items.append(Meta('static_los_troposphere_delay',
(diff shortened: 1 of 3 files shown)


LiangJYu (approved):
LGTM

src/compass/utils/h5_helpers.py
```diff
@@ -617,7 +617,13 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
                  f'Ionospheric delay (range) {desc}',
                  {'units': 'meters'}),
         ]
-        if weather_model_path is not None:
+
+        if weather_model_path is None:
+            correction_items.append(Meta('static_los_troposphere_delay',
+                                             ds.GetRasterBand(5).ReadAsArray(),
+                                             f'Static LOS troposphere delay {desc}',
```
vbrancat, line 624: I would not include the static troposphere as a correction layer inside the product. This correction can be easily calculated using the static layers product. So, there should be no need to have it there too

src/compass/utils/lut.py
```diff
@@ -440,3 +449,27 @@ def resample_set(geo_tide, pts_src, pts_dest):
...
     return rdr_tide
+
+
+def compute_static_troposphere_delay(incidence_angle_arr, hgt_arr):
+    '''
+    Compute troposphere delay using static model
+
+    Parameters:
+    -----------
+    inc_path: str
```
vbrancat, line 460: let's specify that the incidence angle is expected in degrees

vbrancat (commented):
Thanks you @seongsujeong. I am not sure if we should include the static troposphere inside the product. This correction can be easily computed using the static layers product. Let's discuss in more details

tests/test_s1_geocode_slc.py
```diff
@@ -117,4 +117,4 @@ def test_geocode_slc_validate(geocode_slc_params):
 
     # check for bright spots in sliced array
     corner_reflector_threshold = 3e3
-    assert np.any(arr > corner_reflector_threshold)
+    assert np.any(np.abs(arr) > corner_reflector_threshold)
```
seongsujeong, line 120: Here is the suggested change in the test code. 
It looks like `complex64` is being compared to `float`. In this case, it seems numpy only takes the real part of the complex number and to the comparison. I think it makes more sense to calculate the magnitude of the complex image. @LiangJYu What do you think?


> np.array([3001 + 9000 * 1.0j]) > 3e3
array([ True])
> np.array([2001 + 9000 * 1.0j]) > 3e3
array([False])
> np.array([2001 + 1000 * 1.0j]) > 3e3
array([False])
> np.array([3001 + 1000 * 1.0j]) > 3e3
array([ True])

src/compass/utils/h5_helpers.py
seongsujeong: Got it. I've changed the static troposphere LUT not to be written into the product.

src/compass/utils/h5_helpers.py
```diff
@@ -614,16 +614,19 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
         Meta('los_solid_earth_tides', ds.GetRasterBand(4).ReadAsArray(),
              f'Solid Earth tides (range) {desc}',
              {'units': 'meters'}),
+        Meta('los_ionospheric_delay', ds.GetRasterBand(5).ReadAsArray(),
```
seongsujeong, line 617: NOTE: The most recent branch as of now does not seem to have ionosphere LUT written into the product, which seem to happen on #112. Here I add the ionosphere back. Also note that the band ID of the LUTs are adjusted accordingly.

tests/test_s1_geocode_slc.py
LiangJYu: Good find! Please implement your proposed fix.

vbrancat (approved):
Thank you for having addressed my comments. LGTM

PR #101, opened 2023-03-02
CSLC quality assurance

LiangJYu:
This PR adds:
1. computation of min, mean, max of CLSC rasters and corrections
2. determine number of invalid pixels in geocode raster (Very slow!)

To do:

- [x] browse image
- [x] debug slowness or remove feature
- [x] clean up and reorganize

PR #115, opened 2023-03-18
Placeholder for RFI information

seongsujeong:
This PR adds placeholder for RFI information for CSLC output HDF5 file.
The structure of the metadata is based on the pending PR in `s1-reader` for RFI reader [here](https://github.com/opera-adt/s1-reader/pull/97)

### DIFF
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -401,7 +401,7 @@ def metadata_to_h5group(parent_group, burst, cfg):
     # RFI information - placeholder
     rfi_items = [
         Meta('is_rfi_info_available', True,
-             'Whether RFI information is available'),
+             'Boolean variable specifying availability of RFI information'),
         Meta('rfi_mitigation_performed', '',
              'Whether or not the RFI mitigation step was performed'),
         Meta('rfi_mitigation_domain', '',
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -405,7 +405,8 @@ def metadata_to_h5group(parent_group, burst, cfg):
         Meta('rfi_mitigation_performed', '',
              'Whether or not the RFI mitigation step was performed'),
         Meta('rfi_mitigation_domain', '',
-             'in what domain the RFI mitigation step was performed'),
+             ('in what domain the RFI mitigation step was performed '
+              '["Time", "Frequency", "TimeAndFrequency"]')),
         Meta('rfi_burst_report', [],
              'Burst RFI report')]
     rfi_group = meta_group.require_group('rfi_information')
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -398,6 +398,20 @@ def metadata_to_h5group(parent_group, burst, cfg):
     # create metadata group to write datasets to
     meta_group = parent_group.require_group('metadata')
 
+    # RFI information - placeholder
+    rfi_items = [
+        Meta('is_rfi_info_available', True,
+             'Whether RFI information is available'),
+        Meta('rfi_mitigation_performed', '',
+             'Whether or not the RFI mitigation step was performed'),
+        Meta('rfi_mitigation_domain', '',
+             'in what domain the RFI mitigation step was performed'),
+        Meta('rfi_burst_report', [],
+             'Burst RFI report')]
+    rfi_group = meta_group.require_group('rfi_information')
+    for meta_item in rfi_items:
+        add_dataset_and_attrs(rfi_group, meta_item)
+
     # orbit items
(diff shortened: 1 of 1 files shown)


src/compass/utils/h5_helpers.py
```diff
@@ -398,6 +398,20 @@ def metadata_to_h5group(parent_group, burst, cfg):
...
 
+    # RFI information - placeholder
+    rfi_items = [
+        Meta('is_rfi_info_available', True,
+             'Whether RFI information is available'),
+        Meta('rfi_mitigation_performed', '',
+             'Whether or not the RFI mitigation step was performed'),
+        Meta('rfi_mitigation_domain', '',
+             'in what domain the RFI mitigation step was performed'),
+        Meta('rfi_burst_report', [],
```
vbrancat, line 409: What type of information is the `rfi_burst_report` suppose to have? We do process one burst and from the name of the metadata field it seems that there is an info connected to multiple bursts

src/compass/utils/h5_helpers.py
```diff
@@ -398,6 +398,20 @@ def metadata_to_h5group(parent_group, burst, cfg):
     # create metadata group to write datasets to
     meta_group = parent_group.require_group('metadata')
 
+    # RFI information - placeholder
+    rfi_items = [
+        Meta('is_rfi_info_available', True,
+             'Whether RFI information is available'),
```
vbrancat, line 404:
```suggestion
             'Boolean variable specifying availability of RFI information'),
```
Despite I do not know how this metadata field is different from the one below. Maybe historical data do not have RFI info?

src/compass/utils/h5_helpers.py
```diff
@@ -398,6 +398,20 @@ def metadata_to_h5group(parent_group, burst, cfg):
...
+        Meta('rfi_mitigation_performed', '',
+             'Whether or not the RFI mitigation step was performed'),
+        Meta('rfi_mitigation_domain', '',
+             'in what domain the RFI mitigation step was performed'),
```
vbrancat, line 408: `RFI mitigation domain`
Let's specify in parenthesis what these domains are

vbrancat (commented):
Thank you @seongsujeong for this PR. I would like to have the RFI info allocated asa a part of the QA code. Therefore, I am not a fan of their current metadata location. Can we coordinate with @LiangJYu and find a better place for them? Thanks

src/compass/utils/h5_helpers.py
seongsujeong: `rfi_burst_report` would be a list of `dict`, based on the current RFI reader implementation in `s1-reader`.
The reader finds what report corresponds to what burst based on `azimuthTime`, which needs to be reviewed if that is a correct way of doing it.

src/compass/utils/h5_helpers.py
seongsujeong: > Maybe historical data do not have RFI info?
Yes, the RFI information is available from IPF version 3.40. `is_rfi_info_available` is to explicitly indicate that the information is not available.
`rfi_mitigation_performed` is whether RFI mitigation was performed. The existence of this information means that the RFI information is available.

seongsujeong:
> Can we coordinate with @LiangJYu and find a better place for them? Thanks

@LiangJYu Based on #101, I think we write out the QA information into .json file, whose path is `out_paths.stats_json_path`?

LiangJYu:
> > Can we coordinate with @LiangJYu and find a better place for them? Thanks
> 
> @LiangJYu Based on #101, I think we write out the QA information into .json file, whose path is `out_paths.stats_json_path`?

@seongsujeong What do you think about me including the RFI fields in the PR in to #101?

seongsujeong:
>What do you think about me including the RFI fields in the PR in to https://github.com/opera-adt/COMPASS/pull/101?

@LiangJYu Thanks for the suggestion. I think that would be the fastest way to add the placeholder for RFI. The fields for RFI is defined in the list `rfi_items` in `h5_helpers.py` in [here](https://github.com/seongsujeong/COMPASS/blob/6449dda8e02f3d076febfe542b521d0c5e9199c7/src/compass/utils/h5_helpers.py#L401-L411). Please let me know if you need more information.

LiangJYu:
> > What do you think about me including the RFI fields in the PR in to #101?
> 
> @LiangJYu Thanks for the suggestion. I think that would be the fastest way to add the placeholder for RFI. The fields for RFI is defined in the list `rfi_items` in `h5_helpers.py` in [here](https://github.com/seongsujeong/COMPASS/blob/6449dda8e02f3d076febfe542b521d0c5e9199c7/src/compass/utils/h5_helpers.py#L401-L411). Please let me know if you need more information.

Closing this PR. #101 updated to include RFI

PR #117, opened 2023-03-21
x/y coordinate and spacing datasets added to static layer output

LiangJYu:
This PR:
* adds missing x/y coordinates and spacing datasets to the topo static layers
* adds ability to generate topo static layers with `compass.s1_geocode_metadata`
* clean up unused yaml and debug `print`

### DIFF
--- a/src/compass/defaults/s1_cslc_geo.yaml
+++ b/src/compass/defaults/s1_cslc_geo.yaml
@@ -43,7 +43,6 @@ runconfig:
       processing:
           polarization: co-pol
           geocoding:
-              output_format: ENVI
               flatten: True
               lines_per_block: 1000
               x_posting:
--- a/src/compass/s1_cslc_qa.py
+++ b/src/compass/s1_cslc_qa.py
@@ -22,7 +22,6 @@ class QualityAssuranceCSLC:
     Class to compute stats for geocoded raster and corrections
     '''
     stat_names = ['mean', 'min', 'max', 'std']
-    stat_desc = 'Mean, minimum, maximum, and standard deviation of '
 
     def __init__(self):
         self.stats_dict = {}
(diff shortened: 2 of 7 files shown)


LiangJYu:
Now that topo static layers can be generated with `compass.s1_geocode_metadata`, should this ability be removed from `compass.s1_geocode_slc`?

vbrancat:
No, this capability should not be removed from `s1_geocode_slc`. We promised PGE that the static layers would have been generated with the same SAS we use to generate the product. At some point, we want to add the capability to switch on/off the generation of the actual CSLC (to generate the static layers only)

vbrancat (commented):
Thank you @LiangJYu for this PR. Two very broad comments:

1. Can we have the static layer product be named `static_layers_burst_id.h5` instead of `topo.h5`? In this way, we can guarantee that the filename is sort of unique
2. I do not want the static layers product to be too different from the actual CSLC-S1 product.
I am wondering whether we should allocate the static layers under `/science/SENTINEL1/CSLC/grids/static_layers`
3. In the static layer products we should have a burst_id metadata (may be following the same metadata tree of the CSLC-S1 product).

LiangJYu:
> Thank you @LiangJYu for this PR. Two very broad comments:
> 
>     1. Can we have the static layer product be named `static_layers_burst_id.h5` instead of `topo.h5`? In this way, we can guarantee that the filename is sort of unique
> 
>     2. I do not want the static layers product to be too different from the actual CSLC-S1 product.
>        I am wondering whether we should allocate the static layers under `/science/SENTINEL1/CSLC/grids/static_layers`
> 
>     3. In the static layer products we should have a burst_id metadata (may be following the same metadata tree of the CSLC-S1 product).

All 3 points above have been addressed.

vbrancat (approved):
Thanks for having addressed my comments

seongsujeong (approved):
LGTM. Thanks for addressing my comments :)

PR #122, opened 2023-03-22
Use `h5py.Groups.visit()` instead hand rolled code in `utils.validate_cslc._get_group_item_paths()`

LiangJYu:
This PR replace hand rolled recursive HDF5 item getter with [`h5py.Groups.visit()`](https://docs.h5py.org/en/stable/high/group.html#h5py.Group.visit) in `utils.validate_cslc._get_group_item_paths()`.


vbrancat (approved):
LGTM

PR #124, opened 2023-03-23
validation script modified to handle both CSLC and static layer HDF5s

LiangJYu:
This PR modifies `compass.utils.validate_cslc`, renamed to `validate_products`, to handle both CSLC and static layer HDF5s.

Usage:

CSLC: `python -m compass.utils.validate_product -r <ref HDF5> -s <sec HDF5> -p CSLC`

static layer: `python -m compass.utils.validate_product -r <ref HDF5> -s <sec HDF5> -p static_layer`


src/compass/utils/validate_product.py
```diff
@@ -0,0 +1,333 @@
...
+    parser.add_argument(
+        '-s', '--sec-product', type=str, dest='sec_product',
+        help='Secondary CSLC or static layer product to compare with reference')
+    parser.add_argument('-p', '--product-type', type=str, dest='product_type',
+                        choices=['CSLC', 'static_layer'],
+                        default='CSLC', help='Type of file to be validated')
+    return parser.parse_args()
+
+
+def _grid_nfo_retrieve(path_h5, dataset_names, is_static_layer):
```
seongsujeong, line 34: Question: What does `nfo` mean in the function name?

src/compass/utils/validate_product.py
```diff
@@ -0,0 +1,333 @@
...
+        Whether or not path_h5 is static layer product
+
+    Returns
+    -------
+    set
+        Set of names of datasets found in grid group
+    geotransform: np.array
+        Array holding x/y start, x/y spacing, and geogrid dimensions
+    proj: str
+        EPSG projection of geogrid
```
seongsujeong, line 55:
```suggestion
        Map projection of the raster in WKT
```

src/compass/utils/validate_product.py
```diff
@@ -0,0 +1,333 @@
+'''
```
seongsujeong, line 1: What do you think about putting a hashbang, so we can use it as a script? In that case, the permission to execute needs to be added.
```suggestion
#!/usr/bin/env python
'''
```

seongsujeong (approved):
Thanks @LiangJYu for taking care of the validation of the static layer. I've left some suggestion and questions.
the PR looks good to me overall, so I will approve this PR to expedite the process.

src/compass/utils/validate_product.py
LiangJYu: `nfo` was me being lazy and not using `info`. I will correct for clarity.

src/compass/utils/validate_product.py
LiangJYu: Typically COMPASS would need to be installed to run this, so wouldn't it be easier to script with the sample commands I provided in the PR description?

vbrancat (approved):
LGTM

vbrancat:
Can we have the `static_layer` option spelled as `static_layers`

PR #127, opened 2023-03-29
use `isce3::geocode::geocodeSlc` array mode instead of raster mode in `s1_geocode_slc`

LiangJYu:
This PR replaces raster mode in `s1_geocode_slc` with array mode in `isce3::geocode::geocodeSlc`. Array mode includes compression options.

With compression on with default options, a CSLC product shrunk from 1.2G to 147M (this sounds too good to be true). Output was verified correct with visualization in ipython.

~~Needs this [isce3 PR](https://github-fn.jpl.nasa.gov/isce-3/isce/pull/1290) to be merged into develop.~~


src/compass/s1_geocode_metadata.py
```diff
@@ -107,16 +107,19 @@ def run(cfg, burst, fetch_from_scratch=False):
                 continue
 
             dtype = np.single
+            init_value = np.nan
             # layoverShadowMask is last option, no need to change data type
             # and interpolator afterwards
             if layer_name == 'layover_shadow_mask':
                 geocode_obj.data_interpolator = 'NEAREST'
                 dtype = np.byte
+                init_value = 255
```
vbrancat, line 116: Is this the value we attribute to non-valid pixels?

src/compass/s1_geocode_slc.py
```diff
@@ -133,44 +147,74 @@ def run(cfg: GeoRunConfig):
...
+                rdr_datasets.append(rdr_dataset)
+
+                # prepare output dataset in HDF5
+                geo_ds = init_geocoded_dataset(grid_group, pol, geo_grid,
+                                               'complex64',
+                                               f'{pol} geocoded CSLC image',
+                                               np.nan + np.nan * 1j)
+                geo_datasets.append(geo_ds)
+
+            # block proc things
```
vbrancat, line 169: Let's be more descriptive with what is happening here

src/compass/s1_geocode_slc.py
```diff
@@ -133,44 +147,74 @@ def run(cfg: GeoRunConfig):
...
-                gslc_dataset = init_geocoded_dataset(grid_group, pol, geo_grid,
-                                                     'complex64',
-                                                     f'{pol} geocoded CSLC image')
-
-                # Construct the output raster directly from HDF5 dataset
-                geo_burst_raster = isce3.io.Raster(f"IH5:::ID={gslc_dataset.id.id}".encode("utf-8"),
-                                                   update=True)
+                # prepare input dataset as GDAL raster
+                rdr_dataset = gdal.Open(temp_slc_path, gdal.GA_ReadOnly)
+                rdr_datasets.append(rdr_dataset)
```
vbrancat, line 160: Can you comment on what is going on here? So, we are opening GDAL rasters which correspond to the burst we want to process and stack them together in a list? Are we geocoding them together? I am confused because we are still under the `for` loop for the bursts

src/compass/s1_geocode_slc.py
```diff
@@ -133,44 +147,74 @@ def run(cfg: GeoRunConfig):
...
+                                             dest_sel=geo_blk_slice)
+
             # Set geo transformation
+            '''
             geotransform = [geo_grid.start_x, geo_grid.spacing_x, 0,
                             geo_grid.start_y, 0, geo_grid.spacing_y]
             geo_burst_raster.set_geotransform(geotransform)
             geo_burst_raster.set_epsg(epsg)
             del geo_burst_raster
+            '''
```
vbrancat, line 217: How do we assign the geotransformation now that we are not using `isce3.io.Raster`?

src/compass/s1_geocode_slc.py
```diff
@@ -133,44 +147,74 @@ def run(cfg: GeoRunConfig):
...
             del dem_raster # modified in geocodeSlc
```
vbrancat, line 218: Not sure what the comment means. Have we taken care of this at the core module level?

src/compass/utils/h5_helpers.py
```diff
@@ -78,11 +78,13 @@ def add_dataset_and_attrs(group, meta_item):
 
 
 def init_geocoded_dataset(grid_group, dataset_name, geo_grid, dtype,
-                          description):
+                          description, init_value):
     '''
     Create and allocate dataset for isce.geocode.geocode_slc to write to that
     is NC compliant
```
vbrancat, line 84: Confused about the docstring. Can you be more descriptive? Also, do you mean NC stands for NetCDF?

vbrancat (commented):
Thank you for this PR, @LiangJYu . I still need to test this to make sure that this works smoothly. A couple of questions:

1. Are we planning to assign the geotransformation to the generated CSLC? that part of the code (which uses `isce3.io.Raster` seems to be commented out)
2. This PR takes care of generating CSLC in whichever data type we want (e.g. `complex64` or `complex32`)?

src/compass/s1_geocode_metadata.py
LiangJYu: Yes. The value should be 127 which is used in `nisar.workflows.geocode_insar`. I'll add more comments to explain this.

src/compass/s1_geocode_slc.py
LiangJYu: This [first loop](https://github.com/opera-adt/COMPASS/blob/1a8db1f0a365070702560e6466c88fcdd28d243e/src/compass/s1_geocode_slc.py#L139-L167) puts radar and geo datasets into the same place on their respective lists so when they are iterated over independently [before](https://github.com/opera-adt/COMPASS/blob/1a8db1f0a365070702560e6466c88fcdd28d243e/src/compass/s1_geocode_slc.py#L182-L189)/[after](https://github.com/opera-adt/COMPASS/blob/1a8db1f0a365070702560e6466c88fcdd28d243e/src/compass/s1_geocode_slc.py#L204-L208) geocoding block processing the data gets written to the correct place in the HDF5. I will update comments to be more descriptive.

src/compass/s1_geocode_slc.py
LiangJYu: This block is something copied from the NISAR code that I sloppily didn't remove. We set the geotransform in [`compass.utils.h5_helpers.init_geocoded_dataset`](https://github.com/opera-adt/COMPASS/blob/1a8db1f0a365070702560e6466c88fcdd28d243e/src/compass/utils/h5_helpers.py#L80)

src/compass/s1_geocode_slc.py
LiangJYu: In the core module these are buried somewhere in h5_prep.

src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -44,7 +44,8 @@ runconfig:
           polarization: co-pol
           geocoding:
               flatten: True
-              lines_per_block: 1000
+              lines_per_block: 200
+              columns_per_block: 200
```
vbrancat, line 48: bursts are pretty rectangular in radar coordinates, would it make more sense to choose rectangular block sizes? I am not sure whether you have done some analysis to understand what is the impact of block size selection on runtime performances and computational memory usage

vbrancat (commented):
Still some more comments. I did some quick testing and found that in array mode, the CSLC creation for one burst increases by more than one minute (per burst).

The time to geocode one burst in array mode is  2 mins and 58 s (including corrections) while in raster mode is 1 min and 13 s. In array mode, the actual geocoding seems to be very fast but there is some time lag between entering the core module and the actual geocoding of the various data blocks which I was not able to trace back to anything meaningful. Any hint why we spend more time in array mode? Disclaimer: I have used the default block sizes.

LiangJYu:
I benchmarked on my laptop (12 theads) and aurora (112 threads) and put the results [here](https://docs.google.com/spreadsheets/d/1oc4VQ9B0IKsckMUlrvCjxmEhSV6NMOU6sY9umTI-w5k/edit?usp=sharing).

On my laptop, I was able to achieve parity between the array and raster mode by increasing lines/block per column. On aurora, I was able to reduce the difference from ~3x to ~2x by doing the same.

It's unclear at the moment why there's a difference in performance.

src/compass/s1_geocode_slc.py
```diff
@@ -133,44 +149,66 @@ def run(cfg: GeoRunConfig):
...
-                # Construct the output raster directly from HDF5 dataset
-                geo_burst_raster = isce3.io.Raster(f"IH5:::ID={gslc_dataset.id.id}".encode("utf-8"),
-                                                   update=True)
+                # prepare input dataset of current polarization as GDAL raster
+                rdr_dataset = gdal.Open(temp_slc_path, gdal.GA_ReadOnly)
+                rdr_datasets.append(rdr_dataset)
+
+                # prepare output dataset of current polarization in HDF5
+                geo_ds = init_geocoded_dataset(grid_group, pol, geo_grid,
+                                               'complex64',
```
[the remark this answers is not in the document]
LiangJYu: We can add the ability to change this datatype in a future PR.

src/compass/defaults/s1_cslc_geo.yaml
LiangJYu: These 2  parameters are no longer used to geocode the SLC. The S1 bursts are small enough to [read in their entirety into memory](https://github.com/opera-adt/COMPASS/blob/b15bffd6adbb9bb0bf20c3185db5a2c10c9a7c7a/src/compass/s1_geocode_slc.py#L220).

LiangJYu: > These 2 parameters are no longer used to geocode the SLC. The S1 bursts are small enough to [read in their entirety into memory](https://github.com/opera-adt/COMPASS/blob/b15bffd6adbb9bb0bf20c3185db5a2c10c9a7c7a/src/compass/s1_geocode_slc.py#L220).

These 2 parameters are only used for geocoding of metadata.

src/compass/s1_geocode_slc.py
```diff
@@ -167,68 +152,120 @@ def run(cfg: GeoRunConfig):
...
+                                               flatten_phase_ds])
+            geo_data_blks.extend([carrier_phase_data_blk,
+                                              flatten_phase_data_blk])
+            for cslc_dataset, cslc_data_blk in zip(geo_datasets,
+                                                   geo_data_blks):
+                # only convert/modify output if type not 'complex64'
+                # do nothing if type is 'complex64'
+                output_type = cfg.output_params.cslc_data_type
+                if output_type == 'complex32':
+                    cslc_data_blk = nisar.types.to_complex32(cslc_data_blk)
```
scottstanie, line 255:
```suggestion
                    cslc_data_blk = to_complex32(cslc_data_blk)
```
not defined here based on import

src/compass/utils/geo_runconfig.py
```diff
@@ -117,37 +117,41 @@ def load_from_yaml(cls, yaml_runconfig: str, workflow_name: str) -> GeoRunConfig
                    user_plus_default_yaml_str, output_paths, geogrids)
 
     @property
-    def product_group(self) -> dict:
+    def product_group(self) -> types.SimpleNamespace:
```
scottstanie, line 120: I dont think `types` isn't defined in this file anywhere (codacy flagged issue)

src/compass/s1_geocode_slc.py
```diff
@@ -167,68 +152,120 @@ def run(cfg: GeoRunConfig):
...
+                                      first_azimuth_line=0,
+                                      first_range_sample=0,
+                                      flatten=flatten, reramp=True,
+                                      az_carrier=az_carrier_poly2d,
+                                      rg_carrier=isce3.core.Poly2d(np.array([0])),
+                                      az_time_correction=az_lut,
+                                      srange_correction=rg_lut,
+                                      carrier_phase_block=carrier_phase_data_blk,
+                                      flatten_phase_block=flatten_phase_data_blk)
+
```
seongsujeong, line 243:
```suggestion
            flatten_phase_data_blk = (flatten_phase_data_blk + np.pi) % (2*np.pi) - np.pi
            carrier_phase_data_blk = (carrier_phase_data_blk + np.pi) % (2*np.pi) - np.pi
```

@vbrancat The code that I'm suggesting here is to wrap the flattening phases and carrier phases before truncating the mantissa. Once this is committed to this PR, then the geolocation analysis result will get better than truncating the unwrapped phases. I think you do not need to wait until we have the final form of the data optimization in place for your AGE analysis for array mode.

@LiangJYu Can you take a look and apply the changes if it makes sense, so that Virginia can start using the code? Thanks.

requirements.txt
```diff
@@ -2,7 +2,7 @@ python>=3.9     # sentinel1-reader requirement
 numpy           # sentinel1-reader requirement
 lxml            # sentinel1-reader requirement
 gdal>=3
-#isce3        # since the conda-installed isce3 is not the most updated version, installing isce3 from stratch is recommended, to stay in sync with isce3 development.
```
scottstanie, line 5: let's get this pinned like in #170  so that we can correctly install the needed version

src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -46,7 +46,6 @@ runconfig:
           polarization: co-pol
           geocoding:
               flatten: True
-              lines_per_block: 1000
```
scottstanie, line 49: Is this also supposed to be in `rdr2geo` and `geo2rdr` still?

src/compass/utils/h5_helpers.py
```diff
@@ -80,11 +80,13 @@ def add_dataset_and_attrs(group, meta_item):
 
 
 def init_geocoded_dataset(grid_group, dataset_name, geo_grid, dtype,
-                          description, data=None):
+                          description, data=None, output_cfg=None):
     '''
     Create and allocate dataset for isce.geocode.geocode_slc to write to that
-    is NC compliant
+    is NetCDF compliant
```
scottstanie, line 86: small nitpick which you could ignore... i think it's CF-compliant... but not actually "NetCDF compliant" becasue netcdf right now doesnt allow complex data? I could be wrong

src/compass/s1_geocode_slc.py
```diff
@@ -167,68 +157,120 @@ def run(cfg: GeoRunConfig):
...
+                                               'complex64',
+                                               f'{pol} geocoded CSLC image',
+                                               output_cfg=cfg.output_params)
+                geo_datasets.append(geo_ds)
+
+                # Init geocoded output blocks/arrays lists to NaN
+                geo_data_blks.append(
+                    np.full(out_shape, np.nan + 1j * np.nan).astype(np.complex64))
+
+            dt_prep = get_time_delta_str(t_prep)
```
scottstanie, line 206: side note:
```
journal: s1_geocode_slc geocode prep time 0:00:32 (hr:min:sec)
```
this seems kinda slow for just loading data? weird, but we'll look later at it

scottstanie (approved):
I processed a bunch of images with this PR and the interferograms look all good to me. I think I only have tangential comments, and it would be nice to track down what's slowing this down compared to the older version... but if we'd like to get it merged then 👍

src/compass/s1_geocode_metadata.py
```diff
@@ -19,13 +18,15 @@
                                       identity_to_h5group,
                                       init_geocoded_dataset,
                                       metadata_to_h5group, DATA_PATH,
-                                      ROOT_PATH)
-from compass.utils.helpers import bursts_grouping_generator, get_module_name
+                                      METADATA_PATH, ROOT_PATH)
```
seongsujeong, line 21: Can you check if `METADATA_PATH` is used in this code? Thanks

seongsujeong (approved):
Looks good to me. My comments below are just minor ones, so I approve this PR - Seongsu

vbrancat (approved):
LGTM. I agree with @scottstanie that we should investigate why it takes so much time to load the data. Interestingly, the lag he pointed out is consistent with the slowdown I see compared to raster mode.

PR #128, opened 2023-04-01
Enable individual LUTs

vbrancat:


src/compass/utils/lut.py
```diff
@@ -12,242 +12,150 @@
...
+    az_data = np.zeros(lut_shape, dtype=np.float32)
+
+    # Initialize data list and description to save corrections
+    data_dict = {
+        'doppler': [rg_data, 'Slant range geometry and steering doppler'],
+        'bistatic_delay': [rg_data, 'Bistatic delay'],
+        'azimuth_fm_rate': [rg_data, 'Azimuth FM rate mismatch'],
+        'rg_set': [rg_data, 'Slant range Solid Earth Tides'],
+        'az_set': [rg_data, 'Azimuth Solid Earth Tides'],
+        'static_tropo': [rg_data, 'Static troposphere delay'],
```
seongsujeong, line 92:
```suggestion
```
I think we have decided not to include static troposphere correction layer into the output. I've found in `h5_helpers.py` that you indeed skipped writing this layer.
Not sure about why you decided to put the static tropo in to `data_dict` here, but I think it would be less confusing not to include this layer at this point.

src/compass/utils/lut.py
```diff
@@ -12,242 +12,150 @@
...
-        # Compute static troposphere correction
+                                                            rg_step=rg_step)).data
+        az_data -= az_fm_rate
+        data_dict['azimuth_fm_rate'][0] = -az_fm_rate
+
+    # Solid Earth tides
+    if lut_par.solid_earth_tides:
+        dec_factor = int(np.round(5000.0 / rg_step))
+        dec_slice = np.s_[::dec_factor]
+        rg_set_temp, az_set_temp = solid_earth_tides(burst, lat[dec_slice],
```
seongsujeong, line 127: Can you take a look at `lat[dec_slice]` and others to see if its shape is as you intended?  FYI, please take a look at the comment in the link below to see what I've found in SET azimuth PR.

https://github.com/opera-adt/COMPASS/pull/105#discussion_r1136531890

seongsujeong (commented):
Thanks for the PR @vbrancat. Here are my couple comments below
- Seongsu

src/compass/utils/lut.py
seongsujeong: Sounds good. Thanks for the clarification.

src/compass/utils/lut.py
vbrancat: Thank you, I think this has been fixed in the latest commit

src/compass/schemas/s1_cslc_geo.yaml
```diff
@@ -85,17 +85,27 @@ geo2rdr_options:
...
+  azimuth_fm_rate: bool(required=False)
+  # Enable/disable Solid Earth tides correction
+  solid_earth_tides: bool(required=False)
+  # Enable/disable ionosphere TEC correction
+  ionosphere_tec: bool(required=False)
+  # Enable/disable static troposphere correction
+  static_troposphere: bool(required=False)
   # Troposphere delay using weather model
-  troposphere: include('troposphere_options', required=False)
+  weather_model_troposphere: bool(required=False)
```
scottstanie, line 105:
```suggestion
  weather_model_troposphere: include('troposphere_options', required=False)
```

scottstanie:
The `s1_geocode_stack.py` will have to be changed to: https://github.com/vbrancat/COMPASS/blob/single_lut/src/compass/s1_geocode_stack.py#L287 this one currently will make invalid YAMLs by having the single bool... not sure if we really want to expose all 6 on/off options to that script though. That seems excessive, especially if we expect most people will want the corrections (assuming they all improve things)

LiangJYu:
The paths from my changes are a mess. I'll fix them as soon as #148 gets merged into main as I see no point in fixing them now and after another merge with main.

PR #119, opened 2023-03-21
Placeholder for geocoded noise LUT

seongsujeong:
This PR is to add the geocoded noise LUT into the metadata. Currently the noise information is included in the product as radargrid. It is necessary to provide the geocoded LUT, rather than radargrid LUT, to make this information useful.

This addition is basically the repetition of #114 (geocoded calibration LUT), also modification of the noise information's location.

PR #114, opened 2023-03-17
Placeholder for geocoded radiometric calibration parameters

seongsujeong:
This PR adds a placeholders for geocoded LUTs for radiometric calibration.
`gamma` and `sigma_naught` are geocoded and placed into the group `/science/SENTINEL1/CSLC/metadata/calibration_information` as shown in the screenshot below.
Screenshot 2023-03-17 at 00 28 42

PR #130, opened 2023-04-03
geocode calibration parameters and noise LUTs

seongsujeong:
This is a follow-up of #114 #119, which were the placeholders for the geocoded radiometric calibration parameters, and thermal noise LUTs respectively. This loads the calibration parameter LUTs from `Sentinel1BurstSLC`, geocode them, and write that into the output.

### DIFF
--- a/src/compass/s1_geocode_metadata.py
+++ b/src/compass/s1_geocode_metadata.py
@@ -183,8 +183,7 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
     ellipsoid = proj.ellipsoid
     burst_id = str(burst.burst_id)
     geo_grid = cfg.geogrids[burst_id]
-    radar_grid = burst.as_isce3_radargrid()
-
+    
     date_str = burst.sensing_start.strftime("%Y%m%d")
     burst_id_date_key = (burst_id, date_str)
     out_paths = cfg.output_paths[burst_id_date_key]
@@ -241,16 +240,15 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
         # Define the radargrid for LUT interpolation
         # The resultant radargrid will have
         # the very first and the last LUT values be included.
-        radargrid_interp = radar_grid.copy()
+        radargrid_interp = burst.as_isce3_radargrid()
         #radargrid_interp = radargrid_interp.multilook(dec_factor[1], dec_factor[0])
 
         radargrid_interp.width = int(np.ceil(burst.width / dec_factor[0]))
-        range_px_interp_vec = np.linspace(0, burst.width, radargrid_interp.width)
+        range_px_interp_vec = np.linspace(0, burst.width - 1, radargrid_interp.width)
         intv_interp_range = range_px_interp_vec[1] - range_px_interp_vec[0]
 
-
         radargrid_interp.length = int(np.ceil(burst.length / dec_factor[1]))
-        azimuth_px_interp_vec = np.linspace(0, burst.length, radargrid_interp.length)
+        azimuth_px_interp_vec = np.linspace(0, burst.length - 1, radargrid_interp.length)
         intv_interp_azimuth = azimuth_px_interp_vec[1] - azimuth_px_interp_vec[0]
 
         if az_lut_grid is not None:
--- a/src/compass/s1_geocode_metadata.py
+++ b/src/compass/s1_geocode_metadata.py
@@ -20,6 +20,7 @@ from compass.utils.h5_helpers import (init_geocoded_dataset,
                                       ROOT_PATH)
 from compass.utils.helpers import bursts_grouping_generator, get_module_name
 from compass.utils.yaml_argparse import YamlArgparse
+from compass.utils.radar_grid import get_decimated_rdr_grd
 
 
 def run(cfg, burst, fetch_from_scratch=False):
@@ -224,6 +225,16 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
         geo_burst_h5.require_group(dst_group_path)
 
     gdal_envi_driver = gdal.GetDriverByName('ENVI')
+
+    # Define the radargrid for LUT interpolation
+    # The resultant radargrid will have
+    # the very first and the last LUT values be included in the grid.
+    radargrid_interp = get_decimated_rdr_grd(burst.as_isce3_radargrid(),
+                                             dec_factor_x_rng, dec_factor_y_az)
+
+    range_px_interp_vec = np.linspace(0, burst.width - 1, radargrid_interp.width)
+    azimuth_px_interp_vec = np.linspace(0, burst.length - 1, radargrid_interp.length)
+
     for item_name, (rg_lut_grid, rg_lut_val,
                     az_lut_grid, az_lut_val) in item_dict.items():
         # prepare input dataset in output HDF5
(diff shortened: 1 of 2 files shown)


LiangJYu (commented):
Everything looks good. Only curious about decimated radar grid computation method.

src/compass/s1_geocode_metadata.py
```diff
@@ -236,16 +236,43 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
...
+        radargrid_interp.width = int(np.ceil(burst.width / dec_factor))
+        intv_interp_range = (burst.width - 1) / (radargrid_interp.width - 1)
+        range_px_interp_vec = np.arange(0, burst.width, intv_interp_range)
+
+        radargrid_interp.length = int(np.ceil(burst.length / dec_factor))
+        intv_interp_azimuth = (burst.length - 1) / (radargrid_interp.length - 1)
+        azimuth_px_interp_vec = np.arange(0, burst.length, intv_interp_azimuth)
+
+        radargrid_interp.range_pixel_spacing *= intv_interp_range
+        radargrid_interp.prf /= intv_interp_azimuth
```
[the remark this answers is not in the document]
seongsujeong: I wanted to ensure that the very first and the last elements of the LUT be included in the new radargrid. In that case, `multilook()` needs to take `azlooks` and `rglooks` as floating point, rather than integer. So I decided to calculate the new radargrid based on `dec_factor`, which is nominal decimation factor.

I've added a comment to explain why I directly calculated the radargrid, Otherwise even I would be perplexed by myself when I revisit this code after few months ;)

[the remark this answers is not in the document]
LiangJYu: I think a multilooked radargrid generated by will cover the first and last elements of the LUT. When multilooked, the sensing and slant range start will [change](https://github.com/isce-framework/isce3/blob/944eba17f4a5b1c88c6a035c2d58ddd0d4f0709c/cxx/isce3/product/RadarGridParameters.h#L173) but grid size in azimuth (seconds) and slant range (meters) will remain the same since the multilooked PRF and range pixel spacing have been scaled accordingly. This is because the sensing and slant range starts are the center of the first pixel in their respective direction and not the actual leading edge of the pixel. The computations in [`isce3::product::RadarGridParameters::contains()`](https://github.com/isce-framework/isce3/blob/944eba17f4a5b1c88c6a035c2d58ddd0d4f0709c/cxx/isce3/product/RadarGridParameters.cpp#L28) demonstrate that.

I've updated my suggestion above to capture additional commands needed to compute the LUTs.

seongsujeong:
Just did a comparison between multilooked radargrid, and the custom-calculated radargrid. The point here is that the decimated LUT covers the CSLC as much as possible, so that we avoid weird interpolation result on the edges. The screenshots below are the comparison

Screenshot 2023-05-23 at 18 57 12
multi looked radargrid


Screenshot 2023-05-23 at 18 57 17
custom-calculated radargrid


Below are from the same data, but on the lower edge

Screenshot 2023-05-23 at 18 57 48
multi looked radargrid


Screenshot 2023-05-23 at 18 57 52
custom-calculated radargrid

The overall shape of the geocoded LUT is as in the screenshot below. The bluish burst in the back is the amplitude of CSLC, and the `magma` colomapped block is the geocoded LUT.
Screenshot 2023-05-23 at 19 05 49

seongsujeong:
@LiangJYu @vbrancat I think this PR is ready for another round of review. Would you mind taking a look? Thanks.

src/compass/s1_geocode_metadata.py
```diff
@@ -155,7 +156,7 @@ def run(cfg, burst, fetch_from_scratch=False):
 
 
 def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
-                 dec_factor=40):
+                 dec_factor=(20, 5)):
```
LiangJYu, line 159:
```suggestion
                 dec_factor_x_rng=20 . dec_factor_y_az=5):
```
Let's break up for clarity.

src/compass/s1_geocode_metadata.py
```diff
@@ -172,17 +173,17 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
         Path in HDF5 where geocode rasters will be placed
     item_dict: dict
         Dict containing item names and values to be geocoded
-    dec_factor: int
-        Decimation factor to downsample the slant range pixels for LUT
+    dec_factor: tuple
+        Decimation factor to downsample the LUT in
+        range and azimuth direction respectively
```
LiangJYu, line 178:
```suggestion
    dec_factor_x_rg: int
        Decimation factor to downsample the LUT in
        x or range direction
    dec_factor_y_az: int
        Decimation factor to downsample the LUT in
        y or azimuth direction
```
Update to docstring

src/compass/s1_geocode_metadata.py
```diff
@@ -236,16 +237,47 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
...
+        radargrid_interp.width = int(np.ceil(burst.width / dec_factor[0]))
+        range_px_interp_vec = np.linspace(0, burst.width - 1, radargrid_interp.width)
+        intv_interp_range = range_px_interp_vec[1] - range_px_interp_vec[0]
+
+        radargrid_interp.length = int(np.ceil(burst.length / dec_factor[1]))
+        azimuth_px_interp_vec = np.linspace(0, burst.length - 1, radargrid_interp.length)
+        intv_interp_azimuth = azimuth_px_interp_vec[1] - azimuth_px_interp_vec[0]
+
+        radargrid_interp.range_pixel_spacing *= intv_interp_range
+        radargrid_interp.prf /= intv_interp_azimuth
```
LiangJYu, line 253: I think this could be useful elsewhere so let's move this block into function `decimate_radar_grid`  [`src/compass/utils/radar_grid.py`](https://github.com/opera-adt/COMPASS/blob/9a35bed3be8f2dce5ebc525ed45292d74a196761/src/compass/utils/radar_grid.py).

Also let's move this function outside the the `for` loop above. AFAICT the decimated and normal radar grid does not change over the iterated items.

src/compass/s1_geocode_metadata.py
```diff
@@ -272,7 +304,7 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
 
 
 def geocode_calibration_luts(geo_burst_h5, burst, cfg,
-                             dec_factor=40):
+                             dec_factor=(20, 5)):
```
LiangJYu, line 307: Let's breakup `dec_factor` to be consistent with `geocode_luts` above.

src/compass/s1_geocode_metadata.py
```diff
@@ -285,19 +317,36 @@ def geocode_calibration_luts(geo_burst_h5, burst, cfg,
...
+              None,
+              None]
+        }
+    geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict_calibration,
                  dec_factor)
 
 
 def geocode_noise_luts(geo_burst_h5, burst, cfg,
-                       dec_factor=40):
+                       dec_factor=(20, 5)):
```
LiangJYu, line 349: Let's breakup `dec_factor` to be consistent with `geocode_luts` above.

src/compass/s1_geocode_metadata.py
```diff
@@ -221,7 +221,7 @@ def geocode_luts(geo_burst_h5, burst, cfg, dst_group_path, item_dict,
         geo_burst_h5.require_group(dst_group_path)
 
     gdal_envi_driver = gdal.GetDriverByName('ENVI')
-    for item_name, _ in item_dict.items():
+    for item_name, lut_rgaz in item_dict.items():
```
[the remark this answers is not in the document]
seongsujeong: Makes sense. That would make the code easier to read. Thanks for the suggestion!

src/compass/s1_geocode_metadata.py
seongsujeong: Good idea. The radargrid decimation routine is moved to a new function `get_decimated_rdr_grd()` in `radargrid.py`

LiangJYu (approved):
LGTM

PR #131, opened 2023-04-04
Remove is_urgent_observation metadata

vbrancat:
We do not have any urgent observation for CSLC. This metadata entry is inherited from NISAR and probably it does not make sense for production.


LiangJYu (approved):
LGTM

PR #133, opened 2023-04-07
Setup documentation using mkdocs

scottstanie:
you can test it out by installing the extra reqs in `docs/requirements.txt` and then running `mkdocs serve`

image

image

TODO: setup the readthedocs.com stuff so it'll get automatically built (I might need someone's permissions)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,8 +151,6 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
 
-.DS_Store
-
 # unit test files/directories
 tests/data/geo_cslc_s1.yaml
 tests/data/2022-10-16_0000_Rosamond-corner-reflectors.csv
--- a/.gitignore
+++ b/.gitignore
@@ -151,6 +151,8 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
 
+.DS_Store
+
 # unit test files/directories
 tests/data/geo_cslc_s1.yaml
 tests/data/2022-10-16_0000_Rosamond-corner-reflectors.csv
--- a/docs/conda-env.yml
+++ b/docs/conda-env.yml
@@ -1,0 +1,19 @@
+name: compass-docs-env
+channels:
+  - conda-forge
+dependencies:
+  - python[version='>=3.9']
+  - numpy
+  - lxml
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -151,6 +151,8 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
 
+.DS_Store
```
vbrancat, line 154: You probably do not need `.gitignore `

vbrancat (approved):
LGTM :), I am really excited for this new feature (it makes us look cool as the displacement team :P).

.gitignore
scottstanie: I can take it out if you'd like. If you make the docs on Mac, it'll make one of these files:
```
$ find . -name ".DS_Store"
./docs/.DS_Store
```
so we'd just have to make sure people manually remove them or dont add them to git

scottstanie:
i tried removing it from the .gitignore just to bump codacy.

Also i didn't write in the PR header: the setup for the docs now is following recommendations from [Diátaxis documentation framework](https://diataxis.fr/). it's what numpy/other big python projects use to structure docs, and I haven't seen any better suggestions for splitting up documentation pages.

vbrancat:
@LiangJYu and @seongsujeong would be great if one of you has time to look at this PR

docs/conda-env.yml
```diff
@@ -0,0 +1,19 @@
...
+dependencies:
+  - python[version='>=3.9']
+  - numpy
+  - lxml
+  - gdal[version='>=3']
+  - pandas
+  - pyproj
+  - pysolid
+  - pytest
+  - raider
```
LiangJYu, line 13:
```suggestion
  - raider-base
```
IIRC we don't need all of `raider`

Should this also change in `COMPASS/requirements.txt`?

LiangJYu (approved):
LGTM

docs/conda-env.yml
scottstanie: are we removing it entirely from `requirements.txt`?

vbrancat: Actually, I think it can be removed from the requirements. The only way we got it to work is to install `RAiDER` from source and install only the dependencies we need to run the troposphere delay. No need to install it as a `conda `dependency, in my opinion

PR #135, opened 2023-04-10
 Fix license/classifiers metadata in setup.cfg

scottstanie:
If you try to upload this to pypi, its rejected with the error: WARNING  Error during upload. Retry with the --verbose option for more details.
ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
         Invalid value for classifiers.
         Error: Classifier 'License = file : LICENSE' is not a valid classifier.

Fix comes from here:
https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#metadata


vbrancat (approved):
LGTM, thanks :)

LiangJYu (approved):
LGTM

PR #136, opened 2023-04-10
validate_cslc -> validate_product

seongsujeong:
This PR is to update `setup.cfg` so that the validation script can be installed properly. This update is due to the file name change in the PR #124 that extends the validation script to work for static layer files as well as the CSLC output.


vbrancat (approved):
Thank you, @seongsujeong for taking care of this. This item was untracked :'(

PR #137, opened 2023-04-15
save only 6 digits of datetime in orbit metadata

scottstanie:
```
>>> import isce3, datetime
>>> isce3.core.DateTime(datetime.datetime(2022, 11, 20, 16, 16, 43, 152613)).isoformat()

'2022-11-20T16:16:43.152613000'
```

the current orbit description says
> description: b'Reference epoch of the state vectors, format: b'YYYY-MM-DD HH:MM:SS.6f'

but isce's `isoformat` gives 9 digits, so it fails `datetime.datetime.fromisoformat` but would also fail if a user looked at the description to parse it

### DIFF
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -293,7 +293,8 @@ def save_orbit(orbit, orbit_direction, orbit_group):
     orbit_group: h5py.Group
         HDF5 group where orbit parameters will be written
     '''
-    ref_epoch = orbit.reference_epoch.isoformat().replace('T', ' ')
+    # isce isoformat gives 9 decimal places, but python `fromisoformat` wants 6
+    ref_epoch = orbit.reference_epoch.isoformat().replace('T', ' ')[:-3]
     orbit_items = [
         Meta('reference_epoch', ref_epoch, 'Reference epoch of the state vectors',
              {'format': 'YYYY-MM-DD HH:MM:SS.6f'}),


vbrancat (approved):
LGTM. I am assuming that it is sufficient for the downstream workflows to save digits

scottstanie:
> LGTM. I am assuming that it is sufficient for the downstream workflows to save digits

just to check on that... I think there never are 9 digits of precision in the Sentinel orbits:
```
(mapping) staniewi:s1-reader$ grep "\<UTC\>" tests/data/orbits/S1A_OPER_AUX_POEORB_OPOD_20210318T120818_V20200510T225942_20200512T005942.EOF 
      <UTC>UTC=2020-05-10T22:59:42.000000</UTC>
      <UTC>UTC=2020-05-10T22:59:52.000000</UTC>
      <UTC>UTC=2020-05-10T23:00:02.000000</UTC>
      <UTC>UTC=2020-05-10T23:00:12.000000</UTC>
      <UTC>UTC=2020-05-10T23:00:22.000000</UTC>
      <UTC>UTC=2020-05-10T23:00:32.000000</UTC>
...
```
- they seem to give them on the second
- even if they don't, they only provide 6 decimals, so s1reader will never have 9 digits when it forms it https://github.com/opera-adt/s1-reader/blob/7def13120735568e4e5addf842e22d3f7b4ed839/src/s1reader/s1_reader.py#L171

LiangJYu (approved):
LGTM

PR #138, opened 2023-04-18
Add product specification version to CSLC product

vbrancat:
This PR adds the product specification version to the CSLC-S1 product. The version of the product specification document can be assigned from the runconfiguration file and it is exposed at the PGE/ADT interface


scottstanie (approved):

LiangJYu (approved):
LGTM

PR #140, opened 2023-04-19
Make RAiDER a local import

vbrancat:
This PR makes the imports necessary to compute the troposphere correction based on weather model as a local instead of a global import. 
This should avoid to solicit the creation of local log file by the RAiDER module and solve the PGE issue reported in this link https://github.com/nasa/opera-sds-pge/wiki/CSLC-S1-Gamma-Acceptance-Testing-Instructions#known-issues

### DIFF
--- a/src/compass/utils/lut.py
+++ b/src/compass/utils/lut.py
@@ -13,9 +13,6 @@ from compass.utils.geometry_utils import enu2los, en2az
 from compass.utils.iono import ionosphere_delay
 from compass.utils.helpers import open_raster
 from compass.utils.helpers import write_raster
-from RAiDER.delay import tropo_delay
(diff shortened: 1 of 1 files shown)


scottstanie (approved):
Should we add a note in the README setup to the effect of 

```markdown
To compute corrections based on an external weather model, you must additionally install `RAiDER` from conda-forge
```

LiangJYu (approved):
LGTM

vbrancat:
@scottstanie I would say no for the note. We have been having trouble to install `RAiDER` from `conda`. Some of their dependencies are not stable and have conflicts with `isce3`. The only way it worked was to install `RAiDER` from scratch and install only the dependencies needed to run the computation of the tropo delay

PR #141, opened 2023-04-20
miscellaneous fix on `lut.py`

seongsujeong:
This PR is to fix the threshold used in `rdr2geo` object, and unnecessary comma in the function header.


vbrancat (approved):
Thanks @seongsujeong

PR #142, opened 2023-04-24
compute CRs' SNRs and add that into output data frame

seongsujeong:
Along with minor fixes, this PR adds a capability to `age.py` to calculate the SNR of the corner reflector's peak, and add that into output CSV so that the following processes can curl out the detection results whose peak visibility is not strong.

### DIFF
--- a/src/compass/utils/age.py
+++ b/src/compass/utils/age.py
@@ -130,13 +130,13 @@ def run(cslc_file, cr_file, csv_output_file=None, plot_age=False,
 
     # Find peak location for every corner reflector in DataFrame
     for idx, row in cr_df.iterrows():
-        x_peak, y_peak, snr_peak = find_peak(cslc_file, int(row['CR_X_CSLC']),
+        x_peak, y_peak, snr_cr = find_peak(cslc_file, int(row['CR_X_CSLC']),
                                              int(row['CR_Y_CSLC']), pol=pol,
                                              mission_id=mission_id, ovs_factor=ovs_factor,
                                              margin=margin)
         x_peak_vect.append(x_peak)
         y_peak_vect.append(y_peak)
-        cr_snr_vect.append(snr_peak)
+        cr_snr_vect.append(snr_cr)
 
     cr_df['CR_X_CSLC_PEAK'] = x_peak_vect
     cr_df['CR_Y_CSLC_PEAK'] = y_peak_vect
@@ -286,7 +286,7 @@ def find_peak(cslc_file, x_loc, y_loc, mission_id='S1',
     img = arr[upperleft_y:lowerright_y, upperleft_x:lowerright_x]
 
     # Check if the SNR of the peak is above the threshold
-    snr_peak_db = get_snr_cr(img)
+    snr_cr_db = get_snr_cr(img)
 
     # Oversample CSLC subset and get amplitude
     img_ovs = isce3.cal.point_target_info.oversample(
@@ -303,7 +303,7 @@ def find_peak(cslc_file, x_loc, y_loc, mission_id='S1',
     x_cr = x_chip + x_spac / 2 + img_peak_ovs[1] * dx1
     y_cr = y_chip + y_spac / 2 + img_peak_ovs[0] * dy1
 
-    return x_cr, y_cr, snr_peak_db
+    return x_cr, y_cr, snr_cr_db
 
 
 def get_cslc(cslc_file, mission_id='S1', pol='VV') -> np.ndarray :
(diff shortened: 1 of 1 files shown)


src/compass/utils/age.py
```diff
@@ -126,18 +126,21 @@ def run(cslc_file, cr_file, csv_output_file=None, plot_age=False,
...
                                    int(row['CR_Y_CSLC']), pol=pol,
                                    mission_id=mission_id, ovs_factor=ovs_factor,
                                    margin=margin)
         x_peak_vect.append(x_peak)
         y_peak_vect.append(y_peak)
+        snr_peak_vect.append(snr_peak)
 
     cr_df['CR_X_CSLC_PEAK'] = x_peak_vect
     cr_df['CR_Y_CSLC_PEAK'] = y_peak_vect
+    cr_df['CR_SNR_PEAK'] = snr_peak_vect
```
vbrancat, line 143: That might be me but I would call this variable only as "CR_SNR". Adding peak is confusing as this metric is not representative of the peak but is a quality metric for the CR

src/compass/utils/age.py
```diff
@@ -282,6 +285,9 @@ def find_peak(cslc_file, x_loc, y_loc, mission_id='S1',
     # Extract an area around x_loc, y_loc
     img = arr[upperleft_y:lowerright_y, upperleft_x:lowerright_x]
 
+    # Check if the SNR of the peak is above the threshold
+    snr_peak_db = get_snr_peak(img)
```
vbrancat, line 289:
```suggestion
    snr_peak_db = get_snr_cr(img)
```

For the same reason as above

src/compass/utils/age.py
```diff
@@ -493,6 +499,45 @@ def get_cslc_epsg(cslc_file, mission_id='S1', pol='VV'):
     return epsg
 
 
+def get_snr_peak(img: np.ndarray, cutoff_percentile: float=3.0):
+    '''
+    Estimate the signal-to-noise ration (SNR) of the peak
```
vbrancat, line 504:
```suggestion
    Estimate the signal-to-noise ration (SNR) of the corner reflector contained in img
```

src/compass/utils/age.py
```diff
@@ -493,6 +499,45 @@ def get_cslc_epsg(cslc_file, mission_id='S1', pol='VV'):
...
+    in the input image patch
+
+    Parameter
+    ---------
+    img: numpy.ndarray
+        SLC image patch to calculate the SNR
+    cutout: float
```
vbrancat, line 511:
```suggestion
    cutoff_percentile: float
```

src/compass/utils/age.py
```diff
@@ -493,6 +499,45 @@ def get_cslc_epsg(cslc_file, mission_id='S1', pol='VV'):
...
+        Cutout ratio of high and low part of the signal to cutoff
```
vbrancat, line 512: Not sure I fully understand this description here. Can you clarify?

src/compass/utils/age.py
```diff
@@ -493,6 +499,45 @@ def get_cslc_epsg(cslc_file, mission_id='S1', pol='VV'):
...
+
+    Returns
+    -------
+    snr_peak_db: float
+        SNR of the peak in decibel (db)
```
vbrancat, line 517: I would say of the CR

vbrancat (commented):
Some minor comments. I did not have time to test. Hopefully will do it soon

src/compass/utils/age.py
hfattahi: essentially the function is raking a window of the data and returns the ratio of the max power relative to the average power in the window. So it is the SNR of the peak. isn't it?

hfattahi (approved):
LGTM. Thanks. This will make the evaluation easier.

src/compass/utils/age.py
seongsujeong: Makes sense. I'll change the variable names accordingly.

src/compass/utils/age.py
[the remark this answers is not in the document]
seongsujeong: My lack in English skill to use proper preposition lol
Thanks for pointing this out.

src/compass/utils/age.py
seongsujeong: @hfattahi @vbrancat I think both of your suggestion makes sense, but here I personally would prefer to say `snr_cr` because the peak is the "signal" of corner reflectors.

vbrancat (approved):
LGTM :). Disregard my comments

PR #143, opened 2023-04-28
Option to `s1_rdr2geo` to process a single burst

seongsujeong:
This PR adds an option to `s1_rdr2geo` to process a single burst. This is enabled by adding a parameter `burst` to `s1_rdr2geo.run`. If this parameter is not specified, then `s1_rdr2geo.run` will process all bursts in `cfg`.

### DIFF
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -94,7 +94,7 @@ def run(cfg: GeoRunConfig):
 
         # Generate required static layers
         if cfg.rdr2geo_params.enabled:
-            s1_rdr2geo.run(cfg, save_in_scratch=True)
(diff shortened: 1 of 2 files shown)


LiangJYu (commented):
LGTM

vbrancat (approved):
Thank you @seongsujeong LGTM. I tested this PR VS the runtime we have to produce static layers for 3 bursts

Runtime for this PR: 5 min 17 s
Runtime for main branch: 09 min and 15 s

Thank you. This should give a significant improvement to the runtime when generating static layers. @LiangJYu if you are happy with this PR, let's approve it.

LiangJYu (approved):
oops didn't properly approve when I commented :facepalm:

PR #145, opened 2023-05-02
Optimization of function call `make_browse_image`

seongsujeong:
This PR address the issue regarding the runtime when generating browse image. The browse image is generated by using `make_browse_image` in `s1_geocode_slc`. Currently the function takes `cfg.bursts` as input, so that the browse images for all input bursts are generated in each burst for loop. This PR replaces `cfg.bursts` into `bursts` so that the browse image is generated only for the corresponding bursts in the loop.


LiangJYu (approved):
LGTM

vbrancat (approved):
LGTM :)

PR #149, opened 2023-05-06
Change product version and spec version to strings

vbrancat:
This PR changes the following entry of the schema

- product_version
- product_specification_version

to be strings instead of numbers. In this way, we can support versioning with the following format e..g., 0.1.5


LiangJYu (approved):
LGTM

vbrancat:
@seongsujeong can you go through this PR? Thanks

scottstanie (approved):

PR #148, opened 2023-05-06
Remove /science/SENTINEL1 from CSLC-S1 and static layers

vbrancat:
This PR removes the path "s/science/SENTINEL1" to both the CSLC-S1 and the static layers product. 
In addition, the PR adds new metadata entries to the identification group.

The PR has been tested by generating the CSLC product and the static layers for one burst only.

PR #150, opened 2023-05-08
replace real/imag stats with power/phase stats

LiangJYu:
This PR aims to make QA stats for CSLC consistent with that of *SLC in NISAR QA. To expedite the merging of this PR, the stats (min, max, mean, stddev) are computed with `numpy`. The `isce3.math.stats` class with its multithreading capability can not be as easily adapted to apply power and phase to a complex raster.

TODO:
~testing~


LiangJYu:
> @LiangJYu is testing still pending?

Good to go. I think #148 should take precedence over this PR to prevent conflicts there.

LiangJYu:
Need to update specifile.txt with `pytest-order`

src/compass/s1_cslc_qa.py
```diff
@@ -64,40 +77,33 @@ def compute_CSLC_raster_stats(self, cslc_h5py_root, bursts):
...
+            for pwr_phase in ['power', 'phase']:
                 # create dict to store real/imaginary stat items
-                pol_dict[real_imag] = {}
+                pol_dict[pwr_phase] = {}
 
-                # create HDF5 group for real/imaginary stats of current
+                # create HDF5 group for power or phase stats of current
                 # polarization
-                h5_stats_path = f'{QA_PATH}/statistics/data/{pol}/{real_imag}'
+                h5_stats_path = f'{QA_PATH}/statistics/grids/{pol}/{pwr_phase}'
```
vbrancat, line 94: I think we have changed this to be `data` and not `grids`

vbrancat (commented):
@LiangJYu I do not have any major comment on the code. Will approve after having tested

LiangJYu:
retest this please

tests/test_qa.py
```diff
@@ -0,0 +1,21 @@
...
+import numpy as np
+
+from compass.utils.h5_helpers import QA_PATH
+
+def test_qa_power_stats(geocode_slc_params):
+    def _power_test(arr):
+        return arr > 0.0
+
+    def _phase_test(arr):
+        return abs(arr) > np.pi
```
seongsujeong, line 11: I think I'm not sure how the threshold `np.pi` was determined.

tests/test_qa.py
```diff
@@ -0,0 +1,21 @@
...
+
+    # basic sanity checks of mean, min, and max
+    with h5py.File(geocode_slc_params.output_hdf5_path, 'r') as h5_obj:
+        for pwr_phase, test in {'power':_power_test,
+                                'phase':_phase_test}.items():
+            h5_stats_path = f'{QA_PATH}/statistics/data/VV/{pwr_phase}'
+            stat_names = ['mean', 'min', 'max']
+            for stat_name in stat_names:
```
seongsujeong, line 19:
```suggestion
            for stat_name in stat_names:
                print(f'Testing: {stat_name} of {pwr_phase}')
```
What do you think about briefly printing out what array is tested? If something fails in this iteration, the printout messages will make it easier to pinpoint what portions of the statistics have issue.

tests/test_qa.py
```diff
@@ -0,0 +1,21 @@
...
+                stat_val =h5_obj[f'{h5_stats_path}/{stat_name}'][()]
```
seongsujeong, line 20:
```suggestion
                stat_val = h5_obj[f'{h5_stats_path}/{stat_name}'][()]
```

seongsujeong (commented):
Looks good overall. I've left some question about calculating the statistics for periodic value (i.e. phase), as well as minor suggestions - Seongsu

tests/test_qa.py
LiangJYu: [`np.angle`](https://numpy.org/doc/stable/reference/generated/numpy.angle.html) is used to compute phase and it returns (-pi, pi] so nothing should exceed abs(pi).

I will update documentation with explanations.

src/compass/s1_cslc_qa.py
```diff
@@ -11,6 +11,19 @@
...
+    # internal to function to compute min, max, mean, and std dev of power or
+    # phase of SLC array. Default to phase stat computation.
+    if pwr_phase == 'power':
+        post_op_arr = np.abs(arr)**2
+    else:
+        post_op_arr = np.angle(arr)
+
+    return [np_op(post_op_arr)
+            for np_op in [np.nanmean, np.nanmin, np.nanmax,
+                                      np.nanstd]]
```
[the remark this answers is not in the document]
LiangJYu: [`np.angle`](https://numpy.org/doc/stable/reference/generated/numpy.angle.html) is used to compute phase it returns (-pi, pi] so crossing pi will not be an issue. Let me know if I've fully answered your question.

LiangJYu:
These unit tests are elementary. The code should be seen as a placeholder instead of as the final test. I added them with the idea that it's better to have some test than no test. How do you reviewers feel about this approach?

src/compass/s1_cslc_qa.py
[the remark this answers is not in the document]
scottstanie: [Directional statistics](https://en.wikipedia.org/wiki/Directional_statistics) is what you want @seongsujeong 

samples all around the origin have a high circular variance/circular stddev: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.circvar.html#scipy.stats.circvar

```
from scipy import stats

In [5]: def get_samples(mu=0, sigma=1, shape=(1000,)):
   ...:     return mu + sigma*(np.random.randn(*shape) + 1j*np.random.randn(*shape))
   ...:

In [6]: s_origin = get_samples()

In [8]: stats.circmean(np.angle(s_origin)), stats.circstd(np.angle(s_origin))
Out[8]: (0.14590090570201522, 2.933073976147793)
```


if we had a PS point where all of them are scattered around some 100 + 0j, the circular variance is low
```
In [13]: s_ps = get_samples(mu=100)

In [14]: stats.circmean(np.angle(s_ps)), stats.circstd(np.angle(s_ps))
Out[14]: (6.283115405095653, 0.010020883355500842)
```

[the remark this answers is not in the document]
seongsujeong: Thanks @scottstanie! That's what I was wanted to describe. Great to see that scipy has directional statistics implemented.

tests/test_qa.py
seongsujeong: Thanks for the explanation. Then maybe we should switch the inequality like below?
```suggestion
        return abs(arr) <= np.pi
```

seongsujeong:
>These unit tests are elementary. The code should be seen as a placeholder instead of as the final test. I added them with the idea that it's better to have some test than no test. How do you reviewers feel about this approach?

Fine with me. I think we can put the tests with some relexed thresholds, and revisit this to fine-tune the criteria in the tests.

seongsujeong:
> > These unit tests are elementary. The code should be seen as a placeholder instead of as the final test. I added them with the idea that it's better to have some test than no test. How do you reviewers feel about this approach?
> 
> Fine with me. I think we can put the tests with some relexed thresholds, and revisit this to fine-tune the criteria in the tests.

But just letting you know that the test on the mean of phase failed. Below is the code snipped I've added couple of lines like the snipped below  to print out the stats:

```python
print(f'Testing: {stat_name} of {pwr_phase}', end=' ')
stat_val =h5_obj[f'{h5_stats_path}/{stat_name}'][()]
print(stat_val)

```

And here is what the value looks like:

```
Testing: mean of power 2552.5999
Testing: min of power 2.6862714e-05
Testing: max of power 56852844.0
Testing: mean of phase -0.00043806355
```

So apparently the current criteria for the phase mean needs to be adjusted. Please see my comment [above](https://github.com/opera-adt/COMPASS/pull/150#discussion_r1206000491).

src/compass/s1_cslc_qa.py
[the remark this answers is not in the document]
LiangJYu: In light of this [conversation](https://github.com/opera-adt/COMPASS/pull/150#issuecomment-1563508448), @scottstanie @seongsujeong are you two ok with the test as-is and revisiting the implementation of this test after cal-val?

seongsujeong (approved):
LGTM. Thanks for taking care of my comments - Seongsu

vbrancat (approved):
Tested on one bursts and seems to look ok.

PR #156, opened 2023-05-25
Add individual correction metadata to CSLC-S1

vbrancat:
This PR adds some metadata in CSLC-S1 product to indicate which correction has been applied.
Also corrects the snake VS camel case inconsistency in the name of the flattening flags. 

Can you check if the description of the individual correction metadata looks clear?


src/compass/utils/h5_helpers.py
```diff
@@ -567,13 +567,27 @@ def metadata_to_h5group(parent_group, burst, cfg):
...
+        Meta('geometry_doppler_applied', bool(cfg.lut_params.enabled),
+             "If True, geometry steering doppler correction has been applied"),
+        Meta('solid_earth_tides_applied', bool(cfg.lut_params.enabled),
+             "If True, solid Earth tides correction has been applied"),
+        Meta('static_troposphere_applied', bool(cfg.lut_params.enabled),
+             "If True, troposphere correction based on a static model has been applied"),
+        Meta('ionosphere_tec_applied', tec_corr,
+             "If True, ionosphere correction based on TEC data has been applied"),
+        Meta('troposphere_weather_model_applied', tropo_corr,
+             "If True, troposphere correction based on weather model has been applied")
```
LiangJYu, line 590:
```suggestion
        Meta('dry_troposphere_weather_model_applied', tropo_corr,
             "If True, dry troposphere correction based on weather model has been applied")
        Meta('wet_troposphere_weather_model_applied', tropo_corr,
             "If True, wet troposphere correction based on weather model has been applied")
```
To be consistent #128 which separates both dry and wet.

LiangJYu (approved):
LGTM

vbrancat:
@seongsujeong this PR is ready to be reviewed

src/compass/utils/h5_helpers.py
```diff
@@ -567,13 +567,32 @@ def metadata_to_h5group(parent_group, burst, cfg):
...
+        Meta('topographic_flattening_applied', bool(cfg.geocoding_params.flatten),
+             "If True, CSLC-S1 phase has been flatten with respect to topographic height using a DEM"),
+        Meta('bistatic_delay_applied', bool(cfg.lut_params.enabled),
+             "If True, bistatic delay correction has been applied"),
+        Meta('azimuth_fm_rate_applied', bool(cfg.lut_params.enabled),
+             "If True, azimuth FM-rate mismatch correction has been applied"),
+        Meta('geometry_doppler_applied', bool(cfg.lut_params.enabled),
+             "If True, geometry steering doppler correction has been applied"),
+        Meta('los_solid_earth_tides_applied', bool(cfg.lut_params.enabled),
+             "If True, solid Earth tides correction has been applied in slant range direction"),
```
seongsujeong, line 587: Just wondering if we will need to put a placeholder for sth. like `azimuth_solid_earth_tides_applied`

src/compass/utils/h5_helpers.py
```diff
@@ -567,13 +567,32 @@ def metadata_to_h5group(parent_group, burst, cfg):
...
+        Meta('static_troposphere_applied', bool(cfg.lut_params.enabled),
+             "If True, troposphere correction based on a static model has been applied"),
```
seongsujeong, line 589: What is going to happen for static troposphere correction if `cfg.lut_params.enabled` is True, and we apply weather-model based troposphere correction? I think this field has to be `False` in that case?

seongsujeong (approved):
LGTM overall. Here I left couple of questions that you might want to check.
Approving this PR in advance expedite the process because my comments are minor
-- Seongsu

src/compass/utils/h5_helpers.py
vbrancat: It is not a bad idea. Just in case it will be more difficult to change the product

src/compass/utils/h5_helpers.py
vbrancat: My hope is that when we include the individual enabling/disabling of the LUTs, each of these flags will be replaced by its own individual flag. So later on, we should have the ability to set to `False` (or `True`) each of these corrections

PR #132, opened 2023-04-04
Set az conversion

seongsujeong:
This PR is a replacement of #105 which was closed accidentally. 

Most of the comments / suggestions were applied in the previous PR. 

Below is my comment regarding the suggestion to separate `enu2rgaz` into `enu2rg` and `en2az`:
https://github.com/opera-adt/COMPASS/pull/105#discussion_r1157843023

PR #158, opened 2023-05-26
adds azimuth component of SET

seongsujeong:
This PR is the follow-up fix of #132. The issue was that the azimuth component of the SET was computed, but was not added to `az_lut_data`


seongsujeong:
Closing this PR for further test. #159

PR #161, opened 2023-05-28
Force `s1_rdr2geo` to correctly run in `s1_geocode_slc` for static layer generation

LiangJYu:
This PR:
1. forces `s1_rdr2geo` inside `s1_geocode_slc` to generate layers needed for layover shadow computation by creating a new runconfig with the correct flags enabled
2. allow runconfigs to be loaded by yaml string
3. removes unnecessary rdr2geo runconfig check
4. updates defaults so x and y layers are off by default

### DIFF
--- a/src/compass/defaults/s1_cslc_geo.yaml
+++ b/src/compass/defaults/s1_cslc_geo.yaml
@@ -86,7 +86,7 @@ runconfig:
               # Enable/disable computation of longitude raster
               compute_longitude: False
               # Enable/disable computation of height raster
-              compute_height: True
+              compute_height: False
               # Enable/disable layover shadow mask output
               compute_layover_shadow_mask: True
               # Enable/disable incidence angle output
--- a/src/compass/defaults/s1_cslc_geo.yaml
+++ b/src/compass/defaults/s1_cslc_geo.yaml
@@ -82,9 +82,9 @@ runconfig:
               # Secondary number of iterations
               extraiter: 10
               # Enable/disable computation of latitude raster
-              compute_latitude: True
+              compute_latitude: False
               # Enable/disable computation of longitude raster
-              compute_longitude: True
+              compute_longitude: False
               # Enable/disable computation of height raster
               compute_height: True
               # Enable/disable layover shadow mask output
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -564,11 +564,13 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
 
     # Add parameters group in processing information
     if save_processing_parameters:
-        dry_tropo_corr = True if (cfg.weather_model_file is not None) and \
-                                 ('dry' in cfg.tropo_params.delay_type) else False
-        wet_tropo_corr = True if (cfg.weather_model_file is not None) and \
-                                 ('wet' in cfg.tropo_params.delay_type) else False
-        tec_corr = True if cfg.tec_file is not None else False
+        dry_tropo_corr_enabled = \
+            True if (cfg.weather_model_file is not None) and \
+            ('dry' in cfg.tropo_params.delay_type) else False
+        wet_tropo_corr_enabled = \
+            True if (cfg.weather_model_file is not None) and \
+            ('wet' in cfg.tropo_params.delay_type) else False
+        tec_corr_enabled = True if cfg.tec_file is not None else False
         par_meta_items = [
             Meta('ellipsoidal_flattening_applied',
                  bool(cfg.geocoding_params.flatten),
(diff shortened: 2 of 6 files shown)


src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -82,9 +82,9 @@ runconfig:
               # Secondary number of iterations
               extraiter: 10
               # Enable/disable computation of latitude raster
-              compute_latitude: True
+              compute_latitude: False
               # Enable/disable computation of longitude raster
-              compute_longitude: True
+              compute_longitude: False
               # Enable/disable computation of height raster
               compute_height: True
```
vbrancat, line 89: For some reason, I had the height disabled and it was still appearing in my static layers

vbrancat (commented):
Changes look ok for me. However, my static layers product, despite it had the height disabled, it still was contained in the product. Can we fix this?

src/compass/utils/h5_helpers.py
```diff
@@ -560,38 +563,45 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True):
...
-        Meta('dry_troposphere_weather_model_applied', dry_tropo_corr,
-             "If True, dry troposphere correction based on weather model has been applied"),
-        Meta('wet_troposphere_weather_model_applied', wet_tropo_corr,
-             "If True, wet troposphere correction based on weather model has been applied")
-    ]
-    par_meta_group = processing_group.require_group('parameters')
-    for meta_item in par_meta_items:
-        add_dataset_and_attrs(par_meta_group, meta_item)
+    if save_processing_parameters:
+        dry_tropo_corr = True if (cfg.weather_model_file is not None) and \
```
seongsujeong, line 567: `dry_tropo_corr` -> `dry_tropo_corr_enabled`
`wet_tropo_corr` -> `wet_tropo_corr_enabled`
`tec_corr` -> `tec_corr_enabled`

I would like to suggest the variable names above for clarity.

seongsujeong, line 567: `dry_tropo_corr` -> `dry_tropo_corr_enabled`
`wet_tropo_corr` -> `wet_tropo_corr_enabled`
`tec_corr` -> `tec_corr_enabled`

I would like to suggest the variable names above for clarity.

src/compass/utils/runconfig.py
```diff
@@ -48,22 +49,29 @@ def load_validate_yaml(yaml_path: str, workflow_name: str) -> dict:
...
-    else:
-        raise FileNotFoundError(f'Yaml file {yaml_path} not found.')
+    try:
+        if run_config_is_txt:
+            data = yamale.make_data(content=yaml_runconfig,
+                                    parser='ruamel')
+        else:
+            data = yamale.make_data(yaml_runconfig, parser='ruamel')
+    except yamale.YamaleError as yamale_err:
+        err_str = f'Yamale unable to load {workflow_name} runconfig yaml {yaml_runconfig} for validation.'
```
seongsujeong, line 66:
```suggestion
    except yamale.YamaleError as yamale_err:
    yaml_runconfig_info = yaml_runconfig if os.path.exists(yaml_cunfonfig) else 'in the string contents of runconfig`
        err_str = f'Yamale unable to load {workflow_name} runconfig yaml {yaml_runconfig_info} for validation.'

```

In case the the failure of loading the string of runconfig content, the error message will print out whole contents. I think that would be a bit too much for the error message, so suggesting that the failure happened from the contents as a string.

src/compass/utils/runconfig.py
```diff
@@ -48,22 +49,29 @@ def load_validate_yaml(yaml_path: str, workflow_name: str) -> dict:
...
+        error_channel.log(err_str)
+        raise yamale.YamaleError(err_str) from yamale_err
 
     # validate yaml file taken from command line
     try:
         yamale.validate(schema, data)
     except yamale.YamaleError as yamale_err:
-        err_str = f'Validation fail for {workflow_name} runconfig yaml {yaml_path}.'
+        err_str = f'Validation fail for {workflow_name} runconfig yaml {yaml_runconfig}.'
```
seongsujeong, line 74: Same suggestion as above. Maybe we can re-use `yaml_runconfig_info` in the comment above.

seongsujeong (approved):
The logic and the workaround look good to me. Appriving this PR.
I've left a couple of suggestions in the code - Seongsu.

src/compass/defaults/s1_cslc_geo.yaml
LiangJYu: > For some reason, I had the height disabled and it was still appearing in my static layers

@seongsujeong did you encounter the same error as @vbrancat?  I was unable to replicate the error.

@vbrancat were you testing with the latest from this branch?

vbrancat (approved):
LGTM

PR #162, opened 2023-05-28
Docker image preparation and bump COMPASS version for cal/val delivery

seongsujeong:
- SAS Docker images based on develop version of ISCE3
  - Specfile for the Docker image above
- Bump the COMPASS version from `0.1.5` to `0.2.0`

NOTE:
- The Dockerfile used for the last Gamma release is still in place, in case the new version of ISCE3 becomes available
- This is a PR to prepare a CSLC-S1 SAS as a Docker image. ISCE3 is installed by building the source code from the most up-to-date commit in develop branch in internal GitHub.
- The ISCE used in this Docker image will be replaced either tagged commit (preferably mirrored to the public GitHub), or condo package of ISCE3, once either of them becomes available.

TODO:

- [x] Optimize docker image size


seongsujeong:
Note: This is a CSLC-S1 SAS as a Docker image. ISCE3 is installed by building the source code from the most up-to-date commit in `develop` branch in internal GitHub.
The ISCE used in this Docker image will be replaced either tagged commit (preferably mirrored to the public GitHub), or condo package of ISCE3.

docker/Dockerfile.isce3.dev
```diff
@@ -0,0 +1,84 @@
...
+RUN conda create --name "COMPASS" --file /home/compass_user/OPERA/COMPASS/docker/specfile.isce3.dev.txt
+
+SHELL ["conda", "run", "-n", "COMPASS", "/bin/bash", "-c"]
+
+# Install ISCE3
+RUN mkdir -p /home/compass_user/OPERA/ISCE3/install &&\
+    mkdir -p /home/compass_user/OPERA/ISCE3/build &&\
+    cd /home/compass_user/OPERA/ISCE3/build &&\
+    cmake -DCMAKE_INSTALL_PREFIX=/home/compass_user/OPERA/ISCE3/install -DWITH_CUDA=OFF /home/compass_user/OPERA/COMPASS/isce-develop &&\
+    make -j32 VERBOSE=ON &&\
```
LiangJYu, line 48: How do we know there will always be 32 threads? `-j32` is not ideal for anyone not using a 32-thread machine.

I think the `nproc` command could be used here to address this. Before this install ISCE3 command add:
```
ENV N_THREADS=nproc
```
and the `make` command above becomes:
```
make -j${N_THREADS} VERBOSE=ON &&\
```
Sorry untested.

docker/Dockerfile.isce3.dev
```diff
@@ -0,0 +1,84 @@
...
+    rm -rf /home/compass_user/OPERA/ISCE3/build/* &&\
+    cd $CONDA_PREFIX/lib &&\
+    find /home/compass_user/OPERA/ISCE3/install/lib64 -name "*.so*" -exec ln -s {} \; &&\
+    cd `find $CONDA_PREFIX -name "site-packages"|sort|tail -1` &&\
+    find /home/compass_user/OPERA/ISCE3/install/packages -maxdepth 1 -type d|sort|awk 'NR>1'|xargs -I{} ln -s {} &&\
+    rm -rf /home/compass_user/OPERA/ISCE3/build
+
+WORKDIR /home/compass_user/OPERA
+# Install ISCE3
+
```
LiangJYu, line 59:
```suggestion
```
ISCE3 installed above so this is no necessary?

docker/Dockerfile.isce3.dev
```diff
@@ -0,0 +1,84 @@
...
+# installing RAiDER
+RUN mkdir RAiDER &&\
+    git clone https://github.com/dbekaert/RAiDER.git RAiDER &&\
+    cd RAiDER &&\
+    git checkout tags/v0.4.2 &&\
+    cd .. &&\
+    python -m pip install ./RAiDER
+
+
+
```
LiangJYu, line 77:
```suggestion
```
Extra space clean up

docker/Dockerfile.isce3.dev
```diff
@@ -0,0 +1,84 @@
...
+    cd `find $CONDA_PREFIX -name "site-packages"|sort|tail -1` &&\
+    find /home/compass_user/OPERA/ISCE3/install/packages -maxdepth 1 -type d|sort|awk 'NR>1'|xargs -I{} ln -s {} &&\
+    rm -rf /home/compass_user/OPERA/ISCE3/build
+
+WORKDIR /home/compass_user/OPERA
+# Install ISCE3
+
+
+# installing OPERA s1-reader
+RUN curl -sSL https://github.com/opera-adt/s1-reader/archive/refs/tags/v0.1.7.tar.gz -o s1_reader_src.tar.gz &&\
```
LiangJYu, line 62: Out of curiosity -  why is `s1-reader` retrieved with `curl` and not `git clone`+`git checkout` like RAiDER?

LiangJYu (commented):
Mostly cosmetic comments - aside from the `-j` flag used for compiling ISCE3.

docker/Dockerfile.isce3.dev
seongsujeong: Thanks for the suggestion @LiangJYu. I think that is more general than what I've hard-coded here.
One concern is that I do not want to use all CPU threads int he system. Is there a way to pass half or quarter of `nproc`?

docker/Dockerfile.isce3.dev
seongsujeong: I initially tried the same procedure as `s1-reader` when installing `RAiDER`, but it failed due to the reason that was not so clear to me. Maybe I'll have to reproduce the installation error and report that

seongsujeong:
@LiangJYu @vbrancat As you aware,  ISCE3 0.13.0 has become available this morning. The Docker image that I've prepared is based on the conda package of the ISCE3.

Therefore, this PR will discard the `Dockerfile.isce3.dev` based on the on-the-fly ISCE3 build, and focus on updating the specfile for cal/val SAS delivery.

docker/Dockerfile.isce3.dev
seongsujeong: Closing this comment as I've take out the on-the-fly ISCE3 builder

docker/Dockerfile.isce3.dev
seongsujeong: Closing this comment as I've take out the on-the-fly ISCE3 builder. I'll take another look at the Dockerfile based on condo-based ISCE3.

docker/Dockerfile.isce3.dev
seongsujeong: Closing this comment as I've take out the on-the-fly ISCE3 builder.

seongsujeong:
@LiangJYu Can you take a look if the updates in the Docker image is looking good? Thanks.

docker/Dockerfile
```diff
@@ -1,12 +1,14 @@
 FROM oraclelinux:8
 
 LABEL author="OPERA ADT" \
-    description="s1 cslc 0.2 Gamma release" \
-    version="0.2-gamma"
+    description="s1 cslc 0.3 Gamma release" \
+    version="0.3-gamma"
```
LiangJYu, line 5:
```suggestion
    description="s1 cslc 0.3 calval release" \
    version="0.3-calval"
```
Or should it be delta?

LiangJYu (approved):
Just a nomenclature comment. Everything else looks good

docker/Dockerfile
```diff
@@ -1,12 +1,14 @@
...
 
 RUN yum -y update &&\
     yum -y install curl git &&\
-    adduser compass_user
+    adduser compass_user &&\
+    mkdir -p /.config && chmod -R 777 /.config &&\
+    mkdir -p /.cache && chmod -R 777 /.cache
```
LiangJYu, line 11: Just out of curiosity why is this added?

docker/Dockerfile
seongsujeong: The tag here is to clarify that this Docker image is delivered in the cal/val delivery, so I think I prefer to keep it as it is in this PR.

docker/Dockerfile
seongsujeong: While testing Docker image, some dependencies complained about not having the writable directories as above, so I had to create them and give write permission.

LiangJYu: Can you add a comment to capture the above so this knowledge is not lost?

seongsujeong:
A note about CI test failure:

Did a bit of modification on `test_qa.py` to print out the stats written in the HDF5, and found this information:

>Testing: mean of power = 2552.599609375
Testing: min of power = 2.686270818230696e-05
Testing: max of power = 56852836.0
Testing: mean of phase = -0.0004393019189592451
Testing: min of phase = -3.1415927410125732

The CI test failed when testing the min. value of phase. Apparently its absolute value is slightly larger than `np.pi`
The value would have been calculated using `np.nammin()`, which looks okay to me.

As a temporary fix, I will put a tolerance of `1.0e-8` when checking the stats from phase.

PR #164, opened 2023-05-31
bug fixes in `gdal.Open` and `validate_product.py`

LiangJYu:
This PR:
* wraps `gdal.Open` in `compass.utils.helpers.open_raster` with `try/catch` and provide meaningful error message on failure
* fix path bug in `validate_product`
* fixes edge case when `gdal.Open` tries to load a rdr2geo ENVI binary whose 1st two bytes match those of a JPG by using `gdal.OpenEx` after first `try/except`

### DIFF
--- a/src/compass/s1_geocode_slc.py
+++ b/src/compass/s1_geocode_slc.py
@@ -95,15 +95,17 @@ def run(cfg: GeoRunConfig):
     for burst_id, bursts in bursts_grouping_generator(cfg.bursts):
         burst = bursts[0]
 
+        date_str = burst.sensing_start.strftime("%Y%m%d")
+        geo_grid = cfg.geogrids[burst_id]
+
+        info_channel.log(f'Starting geocoding of {burst_id} for {date_str}')
+
         # Reinitialize the dem raster per burst to prevent raster artifacts
         # caused by modification in geocodeSlc
         dem_raster = isce3.io.Raster(cfg.dem)
         epsg = dem_raster.get_epsg()
         proj = isce3.core.make_projection(epsg)
         ellipsoid = proj.ellipsoid
-
-        date_str = burst.sensing_start.strftime("%Y%m%d")
-        geo_grid = cfg.geogrids[burst_id]
 
         # Get output paths for current burst
         burst_id_date_key = (burst_id, date_str)
--- a/src/compass/utils/helpers.py
+++ b/src/compass/utils/helpers.py
@@ -340,14 +340,76 @@ def open_raster(filename, band=1):
     '''
     try:
         ds = gdal.Open(filename, gdal.GA_ReadOnly)
-        raster = ds.GetRasterBand(band).ReadAsArray()
-    except ValueError:
-        error_channel = journal.error('helpers.check_dem')
-        err_str = f'{filename} cannot be opened by GDAL'
+        arr = ds.GetRasterBand(band).ReadAsArray()
+        return arr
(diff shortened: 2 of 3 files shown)


seongsujeong (approved):
LGTM

seongsujeong:
@LiangJYu In the context of providing information debugging, what do you think about printing out the burst ID to the logger? maybe somewhere in the lines below?

https://github.com/opera-adt/COMPASS/blob/643b951efb0f2763b104163365fae7515f52d6f4/src/compass/s1_geocode_slc.py#L95-L98

LiangJYu:
> @LiangJYu In the context of providing information debugging, what do you think about printing out the burst ID to the logger? maybe somewhere in the lines below?
> 
> https://github.com/opera-adt/COMPASS/blob/643b951efb0f2763b104163365fae7515f52d6f4/src/compass/s1_geocode_slc.py#L95-L98

I like the idea. I don't immediately see a clean way of doing it though. Will stew on this...

LiangJYu:
> > @LiangJYu In the context of providing information debugging, what do you think about printing out the burst ID to the logger? maybe somewhere in the lines below?
> > https://github.com/opera-adt/COMPASS/blob/643b951efb0f2763b104163365fae7515f52d6f4/src/compass/s1_geocode_slc.py#L95-L98
> 
> I like the idea. I don't immediately see a clean way of doing it though. Will stew on this...

@seongsujeong rather than passing around the burst ID so `open_raster` what do you think about adding the following call somewhere after the block you linked above?
```python
info_channel.log(f"Starting geocoding of {burst_id} for {date_str}")
```
This keeps `open_raster` general while adding log info that can facilitate log readability.

seongsujeong:
> > > @LiangJYu In the context of providing information debugging, what do you think about printing out the burst ID to the logger? maybe somewhere in the lines below?
> > > https://github.com/opera-adt/COMPASS/blob/643b951efb0f2763b104163365fae7515f52d6f4/src/compass/s1_geocode_slc.py#L95-L98
> > 
> > 
> > I like the idea. I don't immediately see a clean way of doing it though. Will stew on this...
> 
> @seongsujeong rather than passing around the burst ID so `open_raster` what do you think about adding the following call somewhere after the block you linked above?
> 
> ```python
> info_channel.log(f"Starting geocoding of {burst_id} for {date_str}")
> ```
> 
> This keeps `open_raster` general while adding log info that can facilitate log readability.

I like that implementation. Thanks :)

seongsujeong:
Not sure if you like the idea. But if we can exploit this PR for the last fix, can we comment out or remove the couple of lines in `validate_product.py`? I've found that this is necessary for the static layer products from CAL/VAL SAS.

https://github.com/opera-adt/COMPASS/blob/643b951efb0f2763b104163365fae7515f52d6f4/src/compass/utils/validate_product.py#L58-L59

LiangJYu:
> Not sure if you like the idea. But if we can exploit this PR for the last fix, can we comment out or remove the couple of lines in `validate_product.py`? I've found that this is necessary for the static layer products from CAL/VAL SAS.
> 
> https://github.com/opera-adt/COMPASS/blob/643b951efb0f2763b104163365fae7515f52d6f4/src/compass/utils/validate_product.py#L58-L59

Done

vbrancat (approved):
LGTM. My test went fine.

seongsujeong (approved):
LGTM. Thanks for the fix

PR #167, opened 2023-06-02
Correctly populate product fill values in geocoded datasets

LiangJYu:
This PR:
1. correctly populates geocoded dataset fill values based on dataset data type
2. renames `init_geocoded_dataset` to `create_geocoded_dataset`
3. renames `h5py.Dataset` variables with `grid` prefix to `data` prefix


vbrancat:
@LiangJYu do you mind update this branch? So we can go ahead and test it. Thanks.

vbrancat:
@LiangJYu do you mind merging the `main` branch into this PR branch?

src/compass/utils/h5_helpers.py
```diff
@@ -79,6 +79,33 @@ def add_dataset_and_attrs(group, meta_item):
         val_ds.attrs[key] = _as_np_string_if_needed(val)
 
 
+def _determine_fill_value(dtype):
+    '''
+    Helper function to determine COMPASS specific fill values based on h5py
+    Dataset type (dtype)
+    '''
```
vbrancat, line 86: Let's try to have full Numpy docstring with description of inputs and outputs

src/compass/utils/h5_helpers.py
```diff
@@ -79,6 +79,33 @@ def add_dataset_and_attrs(group, meta_item):
...
+    # Possible float types and float fill
+    float_types = [np.double, np.single, np.float32, np.float64, 'float32',
```
vbrancat, line 88: Why do we differentiate between `np.float32` and `float32`?

src/compass/utils/h5_helpers.py
```diff
@@ -79,6 +79,33 @@ def add_dataset_and_attrs(group, meta_item):
...
+                   'float64']
+    float_fill = np.nan
+
+    # Possible complex types and complex fill
+    complex_types = [np.complex128, 'complex64', np.complex64, 'complex32']
+    complex_fill = np.nan * (0 + 1j)
```
vbrancat, line 94: Since this is a tunable parameter in the core module, I am wondering if we can extract it in a more automated way (just in case we change decision on what would be the fill value). Same for other data type. Also, have we made sure that we have been consistent for datasets with the same data type? Please, double-check.

vbrancat (commented):
@LiangJYu, thanks for this PR. I am a bit lost. I see you implemented a functionality to retrieve the fill value for datasets with different data types. However, I don't see anywhere in the code that these fill values are plugged into the product metadata. Can you clarify?

Also, it would be great to come up with an automated way of understanding which fill value has been used for each data type instead of hard-coding them. If this is not possible, can we group all the fill value somewhere more visible?

LiangJYu:
> @LiangJYu, thanks for this PR. I am a bit lost. I see you implemented a functionality to retrieve the fill value for datasets with different data types. However, I don't see anywhere in the code that these fill values are plugged into the product metadata. Can you clarify?
> 
> Also, it would be great to come up with an automated way of understanding which fill value has been used for each data type instead of hard-coding them. If this is not possible, can we group all the fill value somewhere more visible?

All the changes have been merged away against `main`. :sob: I will restore and try to implement you suggestion above.

src/compass/utils/h5_helpers.py
```diff
@@ -79,6 +79,33 @@ def add_dataset_and_attrs(group, meta_item):
         val_ds.attrs[key] = _as_np_string_if_needed(val)
 
 
+def _determine_fill_value(dtype):
```
seongsujeong, line 82: I might be wrong, but I'm not sure where this function is used.
The commit I'm seeing right now is `3af5b7eeaa42c45fb4b71385b3da28c5137fb992`

LiangJYu: It was in 98be0896a034da26990d4dc4243ae2036613cd8b but got merged away. Working to restore.

src/compass/utils/h5_helpers.py
LiangJYu: For its `dtype` parameter, `require_dataset` can take  the string 'float32' and class/type `np.float32`. We use both in our code.

src/compass/utils/h5_helpers.py
LiangJYu: To address your concerns, I've created a new file (`fill_value.py`) that includes:
1. default fill values in a `dataclass`
2. said `dataclass` can handle user defined values
3. `determine_fill_value` function (moved from `h5_helpers.py`)

I'll test more robustly over the weekend.

vbrancat:
@LiangJYu can we quickly address these comments by EOD?

LiangJYu:
@seongsujeong @vbrancat Comments have been addressed. Products look good from my testing with the Peru dataset. Please have a look. Thanks.

vbrancat (commented):
Do you mind merging the `main` branch? It would be much easier for testing. Merci.

vbrancat:
@LiangJYu I have processed a CSLC-S1 product using this PR and inspected it using HDF5. In the HDF5 viewer, I am not able to see the _FillValue as an attribute of 2D datasets. Can we insert as attribute for all the 2D datasets (timing corrections included)? Thanks. 


Screenshot 2023-08-07 at 10 50 15 AM

vbrancat (commented):
Looks good to me but for consistency, we should add fill values also to the timing corrections since these are 2D datasets. @seongsujeong, please, go ahead and review this PR as well.

LiangJYu:
> Looks good to me but for consistency, we should add fill values also to the timing corrections since these are 2D datasets. @seongsujeong, please, go ahead and review this PR as well.

Safe to assume NaN as fill value for all correction values? Apologies, I don't know them off the top of my head.

src/compass/utils/h5_helpers.py
```diff
@@ -120,15 +127,28 @@ def init_geocoded_dataset(grid_group, dataset_name, geo_grid, dtype,
...
+                                             shape=shape, fillvalue=_fill_val,
+                                             **output_kwargs)
     else:
-        cslc_ds = grid_group.create_dataset(dataset_name, data=data,
+        # Create a dataset with provided data
+        cslc_ds = data_group.create_dataset(dataset_name, data=data,
                                             **output_kwargs)
 
     cslc_ds.attrs['description'] = description
+    cslc_ds.attrs['fill_value'] = _fill_val
```
scottstanie, line 151: Any reason to use `fill_value` instead of `_FillValue` like is [described here in the cf conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#missing-data) ?

scottstanie:
Don't wanna slow this down if you're about to try and get it merged, but if you wanted to cut down on some of the code, you might just be able to use the [numpy dtype 'kind'](https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html#numpy-dtype-kind)

```python

In [13]: dts = ['complex64', 'float64', 'float32', 'uint8']

In [14]: for dt in dts:
    ...:     print(dt, ':', np.dtype(dt).kind)
    ...:
```

that's how they do this in h5netcdf:

https://github.com/h5netcdf/h5netcdf/blob/2951cf8ea3174dde5663451484bde0cebaa535d0/h5netcdf/legacyapi.py#L24-L30
where they use a dict keyed on the numpy dtype kind

src/compass/utils/h5_helpers.py
LiangJYu: > Any reason to use `fill_value` instead of `_FillValue` like is [described here in the cf conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#missing-data) ?

I failed to take CF conventions into account. Thank you for pointing me in the right direction.

PR #174, opened 2023-06-29
Add eap correction flag

seongsujeong:
This PR adds a flag `correction_applied_by_sas` into `/metadata/processing_information/parameters/elevation_antenna_pattern_correction` to indicate if the elevation antenna pattern (EAP) correction was applied by SAS (i.e. COMPASS). Also the EAP information is moved from `/metadata/processing_information/timing_corrections` to `/metadata/processing_information/parameter`

PR #175, opened 2023-07-10
add a flag to indicate if elevation antenna pattern (EAP) correction

seongsujeong:
This PR is the same one as #174, but after cleaning the upstream commit history in the source branch.

Below is the description that I've put in the PR #174:
>This PR adds a flag correction_applied_by_sas into /metadata/processing_information/parameters/elevation_antenna_pattern_correction to indicate if the elevation antenna pattern (EAP) correction was applied by SAS (i.e. COMPASS). Also the EAP information is moved from /metadata/processing_information/timing_corrections to /metadata/processing_information/parameter


src/compass/utils/h5_helpers.py
```diff
@@ -821,9 +822,13 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
         add_dataset_and_attrs(extended_coeffs_group, meta_item)
 
     # write out EAP metadata, if present
+    eap_items = [
+        Meta('correction_applied_by_sas', flag_antenna_pattern_by_sas,
+             'Flag to indicate whether SAS applied EAP correction', { 'units': 'N/A'})
```
vbrancat, line 827: Units are not necessary if this is a string or a flag

vbrancat (commented):
@seongsujeong thank you for this PR. I am not sure I fully understand the implementation. To my understanding:

1. If we apply antenna pattern correction, then we allocate a coarse 2D LUT inside `/metadata/processing_information/timing_corrections`, correct? I am assuming this was not done before
2. I am not sure what `correction_applied_by_sas` is and what it should represent. Can you clarify?

Also, I think we should add a metadata field in `metadata/processing_information/parameters` being `is_antenna_pattern_correction_applied` (or whichever is the convention we selected to name metadata)

seongsujeong:
> @seongsujeong thank you for this PR. I am not sure I fully understand the implementation. To my understanding:
> 
> 1. If we apply antenna pattern correction, then we allocate a coarse 2D LUT inside `/metadata/processing_information/timing_corrections`, correct? I am assuming this was not done before

EAP correction is not a timing correction. It is a correction for phase and/or amplitude ( very old data).
The LUT to apply the EAP correction is named `vec_eap_line` in [here](https://github.com/opera-adt/COMPASS/blob/70feb7795abc5812f7b71d0847278a6c75ee88c2/src/compass/utils/elevation_antenna_pattern.py#L29).It is a 1-D LUT with complex numbers, whose length is the same as the number of samples in radargrid. So it's very small.
Currently we do not save it into CSLC product, but I think it shouldn't be tricky. Do you have any suggestion about where the EAP LUT be located?


> 2. I am not sure what `correction_applied_by_sas` is and what it should represent. Can you clarify?
> 
> Also, I think we should add a metadata field in `metadata/processing_information/parameters` being `is_antenna_pattern_correction_applied` (or whichever is the convention we selected to name metadata)

`correction_applied_by_sas` is to clarify that the correction was applied by this SAS (i.e. s1-reader and COMPASS).
ESA has started to apply EAP correction from `IPF > 2.43` by default. Therefore it would be ESA who do the correction in that case. So I think it would be more informative and useful if we clarify who applied the correction (OPERA SAS or ESA) in case we find issue from the correction. Suggestion for the field name is welcomed.
Maybe we can populate the field with the string `COMPASS` or `ESA` rather than storing the flag.

src/compass/s1_geocode_slc.py
```diff
@@ -243,7 +245,8 @@ def run(cfg: GeoRunConfig):
                 corrections_to_h5group(correction_group, burst, cfg, rg_lut, az_lut,
                                        scratch_path,
                                        weather_model_path=cfg.weather_model_file,
-                                       delay_type=cfg.tropo_params.delay_type)
+                                       delay_type=cfg.tropo_params.delay_type,
+                                       flag_antenna_pattern_by_sas=flag_antenna_pattern_by_sas)
```
LiangJYu, line 249:
```suggestion
                                       flag_antenna_pattern_by_sas=check_eap.phase_correction:)
```
More direct and can remove flag variable above.

src/compass/s1_geocode_slc.py
```diff
@@ -175,13 +175,15 @@ def run(cfg: GeoRunConfig):
                 burst.slc_to_vrt_file(temp_slc_path)
 
                 # Apply EAP correction if necessary
+                flag_antenna_pattern_by_sas = False
```
LiangJYu, line 178:
```suggestion
```
Redundant. See named parameter below.

src/compass/s1_geocode_slc.py
```diff
@@ -175,13 +175,15 @@ def run(cfg: GeoRunConfig):
...
                 if check_eap.phase_correction:
                     temp_slc_path_corrected = temp_slc_path.replace('_temp.vrt',
                                                                     '_corrected_temp.rdr')
                     apply_eap_correction(b,
                                          temp_slc_path,
                                          temp_slc_path_corrected,
                                          check_eap)
+                    flag_antenna_pattern_by_sas = True
```
LiangJYu, line 186:
```suggestion
```
Redundant. See named parameter below.

LiangJYu (approved):
LGTM. Just minor comments.

vbrancat:
> > @seongsujeong thank you for this PR. I am not sure I fully understand the implementation. To my understanding:
> > 
> > 1. If we apply antenna pattern correction, then we allocate a coarse 2D LUT inside `/metadata/processing_information/timing_corrections`, correct? I am assuming this was not done before
> 
> EAP correction is not a timing correction. It is a correction for phase and/or amplitude ( very old data). The LUT to apply the EAP correction is named `vec_eap_line` in [here](https://github.com/opera-adt/COMPASS/blob/70feb7795abc5812f7b71d0847278a6c75ee88c2/src/compass/utils/elevation_antenna_pattern.py#L29).It is a 1-D LUT with complex numbers, whose length is the same as the number of samples in radargrid. So it's very small. Currently we do not save it into CSLC product, but I think it shouldn't be tricky. Do you have any suggestion about where the EAP LUT be located?
> 
> > 2. I am not sure what `correction_applied_by_sas` is and what it should represent. Can you clarify?
> > 
> > Also, I think we should add a metadata field in `metadata/processing_information/parameters` being `is_antenna_pattern_correction_applied` (or whichever is the convention we selected to name metadata)
> 
> `correction_applied_by_sas` is to clarify that the correction was applied by this SAS (i.e. s1-reader and COMPASS). ESA has started to apply EAP correction from `IPF > 2.43` by default. Therefore it would be ESA who do the correction in that case. So I think it would be more informative and useful if we clarify who applied the correction (OPERA SAS or ESA) in case we find issue from the correction. Suggestion for the field name is welcomed. Maybe we can populate the field with the string `COMPASS` or `ESA` rather than storing the flag.

@seongsujeong, sorry, my misunderstanding. No need to save the EAP correction in the product but double-check if @hfattahi has a preference on this matter. 

I doubt that people know what SAS means XD. Do we need to differentiate if it us or ESA that performs the antenna pattern correction? It might be transparent to mention that we do it but this is unnecessary nit-picking IMO.

I would just have a flag `is_antenna_pattern_correction_applied` in `metadata/processing_information/parameters`. Will add this as a topic for our CSLC-S1 meeting

seongsujeong:
@vbrancat @hfattahi 
COMPASS will always use EAP-corrected SLC. In case ESA does not correct EAP, we correct the data using our implementation of their algorithm.
That being said, the field `is_antenna_pattern_correction_applied` will always be `True`, so I am not sure how meaningful this would be.
Instead, what I think it would be useful is to keep track of who did the correction. In case we find some issues in EAP correction (e.g. incorrect correction), we will be able to trace back if the problem is coming from our EAP correction implementation, or somewhere else. I'm happy to discuss this in CSLC meeting.

> I doubt that people know what SAS means XD.
Totally agreed. I have a horrible sense of metadata field naming LOL Let's discuss about it in the meeting as well.

seongsujeong:
Looks like the failure in circleCI comes from changes in the location of `types.py` which will be take care of when we have release cut on ISCE3 side.

>ImportError while importing test module '/home/compass_user/OPERA/COMPASS/tests/test_s1_geocode_slc.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniconda3/envs/COMPASS/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_s1_geocode_slc.py:8: in <module>
    from compass import s1_geocode_slc
../../miniconda3/envs/COMPASS/lib/python3.11/site-packages/compass/s1_geocode_slc.py:34: in <module>
    from isce3.core.types import (truncate_mantissa, to_complex32)
E   ModuleNotFoundError: No module named 'isce3.core.types

src/compass/utils/h5_helpers.py
```diff
@@ -608,10 +609,14 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
...
             ('wet' in cfg.tropo_params.delay_type) else False
         tec_corr_enabled = True if cfg.tec_file is not None else False
+        who_applied_eap_correction = 'OPERA' if eap_correction_applied else 'ESA'
         par_meta_items = [
             Meta('ellipsoidal_flattening_applied',
                  bool(cfg.geocoding_params.flatten),
                  "If True, CSLC-S1 phase has been flattened with respect to a zero height ellipsoid"),
+            Meta('elevation_antenna_pattern_correction_applied',
+                 who_applied_eap_correction,
+                 "'OPERA': correction applied by s1-reader and COMPASS. 'ESA': correction was applied by ESA. 'NONE' when the correction was not applied."),
```
vbrancat, line 619:
```suggestion
            Meta('elevation_antenna_pattern_correction_applied',
                 who_applied_eap_correction,
                 "Antenna pattern correction. OPERA: correction applied by OPERA project; ESA: correction applied by ESA; NONE: correction not applied."),
```
Just a quick rewording. Not sure if NONE should be spelled with capital letters. I prefer `None`.

vbrancat (approved):
LGTM. Just a nit. Thanks

src/compass/utils/h5_helpers.py
seongsujeong: Sounds good. In terms of `Antenna pattern` I would prefer to explicitly mention `Elevation` because there seems to be a terminology for the antenna pattern in azimuth direction. See [here](https://www.researchgate.net/figure/Azimuth-antenna-element-pattern-of-Top-Sentinel-1A-EW-mode-and-Bottom-corresponding_fig1_321505761)

PR #176, opened 2023-07-10
WIP CSLC product size optimization

seongsujeong:
This PR is to optimize the product size of CSLC-S1, especially azimuth carrier phase, and flattening phase layers that are currently in full resolution.

At the moment, the simplest thing to apply is to reduce the precision (i.e. float64 -> float32) of the azimuth carrier phase (unwrapped). In case of flattening error (unwrapped), the precision reduction is not suggested because doing it will cause significant truncation error due to its magnitude and range. Detailed information is presented in the PPTX in google drive.

Other optimization algorithm will be added in this PR after discussion.


vbrancat:
Wondering if you tested this approach in combination with the production of CSLC in array mode. Do you see any extra reduction in data size by combining the two?

seongsujeong:
> Wondering if you tested this approach in combination with the production of CSLC in array mode. Do you see any extra reduction in data size by combining the two?

What I've found so far is that `truncate_mantissa()` causes significant truncation error when processing the unwrapped flattening phase and azimuth carrier phase (1st screenshot below). Wrapping and zeroing-out the value has gave reasonable truncation error (2nd screenshot below). 


Screenshot 2023-07-11 at 14 04 05


Screenshot 2023-07-11 at 14 04 13



Regarding the combination: Below are my thoughts about the optimization strategy and/or their combination.
- I don't think precision reduction is not relevant when we apply mantissa truncation.
- sub-sample carrier phase, leave it unwrapped, and `truncate_mantissa()`: The magnitude of the subsampled values will be in the same order of magnitudes for the original signal. Therefore doing it will cause truncation error that mentioned above.
- sub-sample carrier phase, wrap, and truncate_mantissa(): re-sampling this signal will give wrapped phases. We will lose high frequency signals of the phase in that case.
- phase quantization and `truncate_mantissa()`: The idea of mantissa truncation is for floating data.

seongsujeong:
What I see from PR #127 is that we can selectively apply mantissa truncation for each full-res layers. Maybe it's a good idea to implement the optimization algorithm on top of that PR.

vbrancat:
@seongsujeong I have looked at your presentation and understood the conclusions. I think is a good idea starting developing on the array mode. Given the significant reduction in size that we get, it seems that is the way to go.

PR #179, opened 2023-07-17
Wrap the carrier phase and flattening phase right before writing it

seongsujeong:
This PR is an attempt to fix the issue of having empty arrays for azimuth carrier phase and flattening phase. The fix is to retrieve unwrapped phases from `isce3.geocode.geocode_slc`, and wrap if before applying the truncation and writing it into HDF file.


LiangJYu (approved):
LGTM

vbrancat (approved):
LGTM. Thank you @seongsujeong for having caught this.

PR #180, opened 2023-07-18
WIP Data size optimization for azimuth carrier phase and flattening phase

seongsujeong:
This WIP PR is to add the data size optimization algorithm for azimuth carrier phase & flattening phase. Currently they are saved in the same resolution as CSLC layer, which means their sizes can go several hundred megabytes.
The optimization strategy in discussion is mantissa truncation and phase quantization. Currently mantissa truncation was used in this PR to optimize those layers. Phase quantization will be implemented into this PR when the team decided to use that algorithm.

The plot below is the performance comparison between the two algorithms, with variable numbers of significant bits in case of `truncate_mantissa()`

Screenshot 2023-07-13 at 09 14 32


seongsujeong:
Closing this PR as we decided to use mantissa truncation strategy.

PR #188, opened 2023-07-31
Fix the bug in selecting orbit_type

vbrancat:
This causes some SAFE file to fail. This solves the issue.


seongsujeong (approved):
LGTM

scottstanie:
Just curious @vbrancat what was the orbit filename which was had both RESORB and POEORB in it?

PR #190, opened 2023-07-31
Remove spurious flattening metadata from CSLC-S1

vbrancat:
This PR removes extra spurious flattening metadata i.e., `ellipsoidalFlatteningApplied` and `topographicFlatteningApplied ` which were erroneously put in the product and they do not meet the snake case convention we have decided to adopt.

### DIFF
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -465,7 +465,7 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
 
     # Get orbit type
     orbit_file_path = os.path.basename(cfg.orbit_path[0])
-    print(orbit_file_path)
+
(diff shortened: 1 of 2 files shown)


src/compass/utils/h5_helpers.py
```diff
@@ -465,6 +465,7 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
 
     # Get orbit type
     orbit_file_path = os.path.basename(cfg.orbit_path[0])
+    print(orbit_file_path)
```
LiangJYu, line 468:
```suggestion
```
Remove debug `print`?

LiangJYu (approved):
LGTM barring cleanup comment

seongsujeong (approved):
LGTM

PR #191, opened 2023-08-01
Add global attributes to CSLC Static Layers

vbrancat:
This PR adds the global attributes to the CSLC-S1 Static Layers product. In addition, the global attributes of the CSLC-S1 product are modified according to the feedback received by SDS.


LiangJYu (approved):
LGTM

seongsujeong (approved):
LGTM

PR #193, opened 2023-08-01
s1_geocode_stack.py: Add ability to read "DH" SAFE files

scottstanie:
This adds the ability to recognize SAFE files that have HH/HV instead of VV/VH. 
Right now I'm raising a `NotImplementedError` in `_get_pol_str` if you try to ask for `dual-pol`, since I'm not sure how we're handling that elsewhere in the code and it seems like it would need some further refactoring to allow.

### DIFF
--- a/src/compass/s1_geocode_stack.py
+++ b/src/compass/s1_geocode_stack.py
@@ -79,7 +79,8 @@ def create_parser():
 
 
 def generate_burst_map(zip_files, orbit_dir, output_epsg=None, bbox=None,
-                       bbox_epsg=4326, burst_db_file=DEFAULT_BURST_DB_FILE):
+                       bbox_epsg=4326, burst_db_file=DEFAULT_BURST_DB_FILE,
+                       pol_type: str = "co-pol"):
     """Generates a dataframe of geogrid infos for each burst ID in `zip_files`.
 
     Parameters
@@ -98,6 +99,8 @@ def generate_burst_map(zip_files, orbit_dir, output_epsg=None, bbox=None,
         to be lon/lat degrees (default: 4326).
     burst_db_file: str
         Path to the burst database file to load bounding boxes.
+    pol-type: str
+        Polarizations to process. Choices: co-pol, cross-pol, dual-pol
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


src/compass/s1_geocode_stack.py
```diff
@@ -135,6 +139,30 @@ def generate_burst_map(zip_files, orbit_dir, output_epsg=None, bbox=None,
...
+        if h_or_v == "H":
+            return ["hh", "hv"]
+        else:
+            return ["vv", "vh"]
+    elif pol_type == "cross-pol":
+        return 'hv' if h_or_v == "H" else "vh"
+    elif pol_type == "co-pol":
+        return 'hh' if h_or_v == "H" else "vv"
+    else:
+        raise ValueError(f"Invalid {pol_type = }")
```
seongsujeong, line 165:
```suggestion
    if pol_type not in ["dual-pol", "cross-pol", "co-pol"]:
        raise ValueError(f"Invalid {pol_type = }")
    if count == "S" and pol_type in ["dual-pol", "cross-pol"]:
        raise ValueError(f"{zip_file} is single-pol. Unable to process {pol_type}")
    if pol_type == "dual-pol":
        raise NotImplementedError("TODO")
        if h_or_v == "H":
            return ["hh", "hv"]
        else:
            return ["vv", "vh"]
    if pol_type == "cross-pol":
        return "hv" if h_or_v == "H" else "vh"
    if pol_type == "co-pol":
        return "hh" if h_or_v == "H" else "vv"
```

Early returns for readability

seongsujeong:
Another thought. Currently the default pol for `load_bursts()` in `s1-reader` is fixed to `vv` [(here)](https://github.com/seongsujeong/s1-reader/blob/752c711d9fe5fe452c13408e12d163f837f6a81e/src/s1reader/s1_reader.py#L966). I think this functionality would be useful for the s1-reader. So, what do you think about putting this function into `s1-reader` @scottstanie?

scottstanie:
Oh that's a great point. I knew there was somewhere that made more sense than tucked away in this stack script lol.

PR #198, opened 2023-08-11
fix `s1_geocode_stack.py`, undefined variable in `s1_rdr2geo.py`, unit tests

scottstanie:
Also
- Adds a smoke test which runs the static layers script
- add unit test for geocode stack script


scottstanie:
@LiangJYu 
```
tests/test_s1_geocode_stack.py F                                         [ 85%]
tests/test_s1_static_layers.py 
CondaError: KeyboardInterrupt


Too long with no output (exceeded 10m0s): context deadline exceeded
```

the geocode tests took too long, and I don't have permissions to restart on CircleCI

src/compass/s1_static_layers.py
```diff
@@ -75,13 +74,16 @@ def run(cfg: GeoRunConfig):
...
 
     # Get a runconfig dict from command line arguments
     cfg = GeoRunConfig.load_from_yaml(parser.run_config_path,
                                       workflow_name='s1_cslc_geo')
 
-    # Run geocode burst workflow
     run(cfg)
+
+if __name__ == "__main__":
+    main()
```
LiangJYu, line 89: TIL https://docs.python.org/3/library/__main__.html#idiomatic-usage

LiangJYu (commented):
Regarding timeout, @yunjunz long ago had a suggestion that we bump our timeout limit. Based on his suggestion in slack, we can add [this](https://github.com/insarlab/MintPy/blob/99ae8ecdf373dad35fe32a1ec49afce51d4d1bb5/.circleci/config.yml#L37) below `name` [here](https://github.com/opera-adt/COMPASS/blob/a3234717681b81640b3d7e6033120a37055a03e6/.circleci/config.yml#L15).

scottstanie:
image
fixed by making the command `pytest --capture=tee-sys`

LiangJYu (approved):
LGTM

scottstanie:
@vbrancat @seongsujeong I just added a version bump to 0.4.1 to this so that we can have a new conda-forge version with a working `s1_geocode_stack.py`. let me know if there's any problem with that

src/compass/s1_static_layers.py
vbrancat, line 89: Missing blank line at the end of the function?

vbrancat (approved):
LGTM. Thanks for these changes :)

PR #200, opened 2023-08-18
Fix inconsistencies between generated product and product specifications

vbrancat:
This PR resolves the metadata inconsistencies between the generated CSLC-S1 and static layers products with the corresponding product specification documents.

Particularly the PR applies the following fixes:
1. `y_spacing` will be a positive value
2. added `azimuth_solid_earth_tides` to the QA stats group
3. Add description to `projection` metadata field.
4. Correct units of `azimuth_steering_rate`
5. Correct `orbit_type` description

- [Testing and check that the above has been addressed according to the received comments ]


jshimada47:
If the Y-direction is positive going north, shouldn't _y_spacing_ be negative if the records are ordered from north to south?

vbrancat:
@jshimada47 totally agree with that. However, comments from our CSLC-S1 product review mention that a negative spacing is not intuitive for users :s. I am not sure where the confusion is, frankly

jshimada47:
Yes, I don't think that it's unusual for geocoded products to have negative y-spacings because most of them define the positive direction as increasing latitude. I'm concerned that using positive-valued y-spacings contradicts the definition in the product spec.

LiangJYu (approved):
LGTM

vbrancat:
@seongsujeong can you please have a look at this PR? Thanks.

scottstanie (approved):

PR #201, opened 2023-08-22
Add field for DEM description into runconfig and product

seongsujeong:
This PR adds a field for DEM description into runconfig and the output product.
The field `dem_description` will be populated into `/metadata/processing_information/inputs/dem_source`, which was previously being populated with the filename of the DEM i.e. `dem_file`.

### DIFF
--- a/src/compass/defaults/s1_cslc_geo.yaml
+++ b/src/compass/defaults/s1_cslc_geo.yaml
@@ -17,6 +17,8 @@ runconfig:
       dynamic_ancillary_file_group:
           # Digital elevation model
           dem_file:
+          # Description for DEM used for the processing
+          dem_description: DEM description was not provided.
           # TEC file in IONEX format for ionosphere correction
           tec_file:
           # Troposphere weather model file
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -519,7 +519,7 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
              'List of input noise files used for processing'),
         Meta('dem_source',
              os.path.basename(cfg.groups.dynamic_ancillary_file_group.dem_description),
-             'Description for source DEM file'),
+             'Description of the DEM used for processing'),
     ]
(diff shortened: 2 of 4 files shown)


src/compass/defaults/s1_cslc_geo.yaml
```diff
@@ -17,6 +17,8 @@ runconfig:
       dynamic_ancillary_file_group:
           # Digital elevation model
           dem_file:
+          # Description for DEM used for the processing
+          dem_description: DEM description was not provided.
```
vbrancat, line 21: Have you tested this? Shouldn't this be the following:

```suggestion
          dem_description: 'DEM description was not provided.'
```

src/compass/utils/h5_helpers.py
```diff
@@ -510,13 +510,16 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
...
+             'List of input L1 RSLC files used for processing'),
         Meta('orbit_files', orbit_files, 'List of input orbit files used for processing'),
         Meta('calibration_files', burst.burst_calibration.basename_cads,
              'List of input calibration files used for processing'),
         Meta('noise_files', burst.burst_noise.basename_nads,
              'List of input noise files used for processing'),
-        Meta('dem_source', os.path.basename(cfg.dem), 'source DEM file'),
+        Meta('dem_source',
+             os.path.basename(cfg.groups.dynamic_ancillary_file_group.dem_description),
+             'Description for source DEM file'),
```
vbrancat, line 522: Suggesting to change it as:
```suggestion
             'Description of the DEM used for processing'
```

vbrancat (approved):
LGTM. Merging is pending my comments.

src/compass/defaults/s1_cslc_geo.yaml
seongsujeong: confirmed that it works without the quotation marks.
Screenshot 2023-08-21 at 19 01 22

LiangJYu (approved):
LGTM

PR #202, opened 2023-08-22
add `conda clean -afy` after install for smaller image

scottstanie:
drops size by 1 GB


seongsujeong (approved):
LGTM. Thank you for the PR!

PR #196, opened 2023-08-04
Update the environment files and install instructions

scottstanie:

PR #205, opened 2023-08-25
Preparation for final delivery

seongsujeong:
This PR is to prepare for the final delivery of CSLC-S1 SAS.

- [x] Update `Dockerfile`, `environments.yml`, and `specfile.txt` to prepare for the Docker image. Note that s1reader is being installed from conda forge as it became available thanks to @scottstanie.
- [x] Fix the bug found during the test
- [x] Bump the version


docker/env_compass.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - dem_stitcher
+  - gxx_linux-64>=9,<10
+  - pysolid
+  - pytest
+  - pytest-order
+
+
+
+./CSLC_CALVAL_DELIVERY.playground/run_mem_profiler.sh
+./CSLC_CALVAL_DELIVERY.playground/start_docker_container_for_mem_profile.sh
```
vbrancat, line 16: Should this be changed to Final?

environment.yml
```diff
@@ -14,6 +14,17 @@ dependencies:
...
+# CI unit test
+  - pytest
+  - pluggy
+  - pytest-order
+# Dependencies for RAiDER
+  - progressbar
+  - rasterio
+  - dem_stitcher
+  - gxx_linux-64>=9,<10
+  - pysolid
```
vbrancat, line 30: Missing space at the end of file.

vbrancat (approved):
LGTM. Merge pending two minor comments

docker/env_compass.yaml
seongsujeong: This file `env_compass.yaml` was pushed by accident. I'll retract this file. Thanks for pointing this.

seongsujeong:
@vbrancat Thanks for the quick turnaround. I'll merge this after making sure that it passes  the CI test.

src/compass/version.py
```diff
@@ -8,6 +8,7 @@
 # release history
 Tag = collections.namedtuple('Tag', 'version date')
 release_history = (
+    Tag('0.5.0', '2023-08-25'),
```
seongsujeong, line 11: @vbrancat I originally intended this PR to take care of the Docker only, but I think it would be more efficient to to extend its scope to prepare for the final delivery. As part of the preparation, I also included the change to bump the version. Please let me know if you don'y like the version number. Thanks.

vbrancat: I have no problem with the version number.

environment.yml
```diff
@@ -14,6 +14,18 @@ dependencies:
...
   - yamale>=4.0
   - h5py>=3.5
-  - s1reader>=0.2.0
+  - s1reader>=0.2.1
   - shapely>=1.8
   - requests>=2.0
+  - scikit-image
+# CI unit test
+  - pytest
+  - pluggy
```
scottstanie, line 23: Hmm this one doesn't look like it's being used by COMPASS? I think `pluggy` is the module that `pytest` created as a general framework for making plugins

environment.yml
```diff
@@ -14,6 +14,18 @@ dependencies:
...
+  - pytest-order
+# Dependencies for RAiDER
+  - progressbar
+  - rasterio
+  - dem_stitcher
+  - gxx_linux-64>=9,<10
+  - pysolid
```
scottstanie, line 30: We've already got `pysolid` on here as a compass dependency

environment.yml
```diff
@@ -14,6 +14,18 @@ dependencies:
...
   - shapely>=1.8
   - requests>=2.0
+  - scikit-image
+# CI unit test
+  - pytest
+  - pluggy
+  - pytest-order
+# Dependencies for RAiDER
+  - progressbar
+  - rasterio
```
scottstanie, line 27: you've started to use `rasterio` in the COMPASS code here https://github.com/opera-adt/COMPASS/blob/main/src/compass/s1_cslc_qa.py#L14-L15

(As a side note- if the test for the PR that added that is passing... does that mean we should add some [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)) where we're actually building and importing everything? Not sure how that one could have passed)

scottstanie: we also now are importing `geopandas` but don't have it listed here

environment.yml
```diff
@@ -14,6 +14,18 @@ dependencies:
...
-  - s1reader>=0.2.0
+  - s1reader>=0.2.1
   - shapely>=1.8
   - requests>=2.0
+  - scikit-image
+# CI unit test
+  - pytest
+  - pluggy
+  - pytest-order
+# Dependencies for RAiDER
```
scottstanie, line 25: Are we making `raider` a full requirement for COMPASS? 
we seem to have added a lot of it's dependencies now, and keeping this section in the main `environment.yml` file sort of implies we want Raider as a dependency... but we're not specifying `raider` or `raider-base`

src/compass/utils/h5_helpers.py
```diff
@@ -781,7 +781,7 @@ def corrections_to_h5group(parent_group, burst, cfg, rg_lut, az_lut,
              'spacing of slant range of LUT data', {'units': 'meters'}),
         Meta('zero_doppler_time', azimuth, 'azimuth time of LUT data',
              {'units': 'seconds'}),
-        Meta('zero_doppler_time_spacing', np.abs(rg_lut.y_spacing),
```
scottstanie, line 784: was this removed before we don't mind it being negative?

src/compass/s1_cslc_qa.py
```diff
@@ -659,4 +659,5 @@ def _get_land_mask(epsg_cslc: int, geotransform: tuple, shape_mask: tuple):
                       transform=tform_bbox,
                       out_shape=shape_mask,
                       dtype=np.uint8, default_value=1)
+
```
scottstanie, line 662: I can't comment directly on the code above for this PR since it wasn't changed:


```python

    # Make sure the CRS of the bounding box is the same as the GeoDataFrame
    bbox = gpd.GeoSeries([bbox], crs=f'EPSG:{epsg_cslc}')
    if epsg_cslc != 3413:
        bbox_3413 = bbox.to_crs('EPSG:3413')
```
this creates an undefined variable when `epsg_cslc == 3413`


and less, critical, but making a `GeoSeries` with one shapely polygon seems unnecessary since we're just re-selecting the polygon to do `.intersection`

scottstanie:
two things-
1. If I make a lockfile from the environment here, and I compare it to the one by deleting the `Testing` and `Raider` sections, it goes from 261 lines to 201 lines (so 60 fewer dependencies).
The biggest one is that `geopandas` is used in the QA script, imported at the top level, so it's unavoidable right now it's unavoidable. it would be nice to not have to install all of geopandas (which has a huge dependency tree, e.g. `scitkit-learn`) just to read the shapefile. Here's a demo of using shapely/fiona instead: https://github.com/seongsujeong/COMPASS/compare/docker_final_delivery...scottstanie:COMPASS:docker_final_delivery_shapely?expand=1

2. What if we put the raider depencies in a separate file and called `--file ...` twice on the other one to keep it separate? We could then also delete the `testing` section of the `environment.yml` file because [we already have a separate testing requirements file](https://github.com/opera-adt/COMPASS/blob/main/tests/requirements.txt) so we could just reference that (though I dont think we need to install testing dependencies into the main docker anyway?)

vbrancat:
> 2\. he other one to keep i

Totally agree on the dependency of `geopandas`. I expressed my concerns on it in private conversations with @seongsujeong. I have provided Python code on how to open a shapefile with `ogr`. @seongsujeong let's work on avoiding adding this dependency.

environment.yml
seongsujeong: We need those dependency to make the unit test to run in specific order. So we need to keep it as part of the dependency.

environment.yml
seongsujeong: That's True. Will remove the redundant `pysolid` here. Thanks for catching this.

environment.yml
seongsujeong: Based on what I experienced in beta release, installing RAiDER from conda-forge takes forever, and it adds way too much dependencies. So I am not a fan of adding `RAiDER` to `environments.yml`.

src/compass/utils/h5_helpers.py
seongsujeong: That was the request from @vbrancat to revise the code to have `y_spacing` be negative.

seongsujeong:
> two things-
> 
> 1. If I make a lockfile from the environment here, and I compare it to the one by deleting the `Testing` and `Raider` sections, it goes from 261 lines to 201 lines (so 60 fewer dependencies).

  That is expected. And I hope you have made sure that you have made sure that CI passed with the lockfile (i.e. not the YAML file) you have without `Testing` and `RaiDER`, especially without `Testing`.
In the PR #196, you have updated `environment.yml`. I appreciate your effort to make COMPASS as lightweight as possible, but we need an environment that makes COMPASS works, including the unit test. You said it's fine, and I initially agreed that it should be okay, because it passed the CI, but please note that the dependency during the CI gets installed by using `specfile.txt`. So CI pass on that PR does not mean `environments'yaml` is okay, unless you update the specfile based on the new yaml file.

>    The biggest one is that `geopandas` is used in the QA script, imported at the top level, so it's unavoidable right now if you want to run `s1_geocode_slc.py`. it would be nice to not have to install all of geopandas (which has a huge dependency tree, e.g. `scitkit-learn`) just to read the shapefile. Here's a demo of using shapely/fiona instead: https://github.com/seongsujeong/COMPASS/compare/docker_final_delivery...scottstanie:COMPASS:docker_final_delivery_shapely?expand=1
After the discussion with @vbrancat , I decided to write shapely-based anyhow. Doing that will get rid of the necessity of `geopandas`. But please note that `geopandas` is a dependency of RaiDER's dependency, and we need RAiDER for the final SAS delivery. I cannot guarantee that `RAiDER` works without `geopandas`.

> 2. What if we put the raider depencies in a separate file and called `--file ...` twice on the other one to keep it separate? We could then also delete the `testing` section of the `environment.yml` file because [we already have a separate testing requirements file](https://github.com/opera-adt/COMPASS/blob/main/tests/requirements.txt) so we could just reference that (though I dont think we need to install testing dependencies into the main docker anyway?)

Not sure if you mean to have a separate `.yml` file or `lockfile`, but in either case, I am not sure if conda will attempt to resolve the dependency, which I don't want. The `.yml` for sure, The `lockfile` I'm not sure. 

Please, note that we need to focus on two things as ADT: to deliver the SAS in the spec we promised, and to make CI run correctly to deliver SAS that is robust enough. Optimizing the Docker size is a great thing we can do before the delivery, and I again thank you for your efforts, suggestions, and tips. They are really helpful. However, I am going to focus more on the first two prongs rather than the latter one.

About the implementations I've recently put in `s1_cslc_qa.py`, I will spend some time to re-write it using shapely. I'll either submit another PR, or include that in this one.

seongsujeong:
@LiangJYu I've changed the implementation of the land mask computation to use osgeo modules (i.e. ogr, osr, and gdal). Hope you don't mint taking a look at the new code. Than you :)

vbrancat (commented):
New implementation of pixel classification looks good to me

LiangJYu (approved):
LGTM. Pending my minor comments

environment.yml
seongsujeong: Unfortunately, RAiDER still needs them. `geopandas` is not directly required by RaiDER, but I've found that installing the RAiDER dependencies (i.e. `progressbar`, `rasterio`, `dem_stitcher`, and `gxx_linux-64`) resulted in installing `geopandas`.

LiangJYu: My bad - I missed the `RAiDER` comment above

environment.yml
```diff
@@ -14,6 +14,17 @@ dependencies:
...
+  - progressbar
+  - rasterio
+  - dem_stitcher
```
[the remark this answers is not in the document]
seongsujeong: Yes, it was a dependency for RAiDER. I've found that by installing its dependency one by one, and test-running COMPASS.

src/compass/s1_cslc_qa.py
```diff
@@ -620,43 +618,88 @@ def _get_valid_pixel_mask(arr_cslc):
...
-
-    # Check if the polygon intersects with the bounding box.
-    if not gdf1.geometry.iloc[0].intersects(bbox_3413.iloc[0]):
-        # no overlap
+    bbox_cslc = ogr.Geometry(ogr.wkbPolygon)
+    ring_cslc = ogr.Geometry(ogr.wkbLinearRing)
+    ring_cslc.AddPoint(xmin, ymin)
+    ring_cslc.AddPoint(xmax, ymin)
+    ring_cslc.AddPoint(xmax, ymax)
+    ring_cslc.AddPoint(xmin, ymax)
```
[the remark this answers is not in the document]
seongsujeong: Thanks for the suggestion. I've tested the code, but got the result as below:
```
>>> for y, x, in itertools.product(['ymin', 'ymax'], ['xmin', 'xmax']):
...   print(x,y)
... 
xmin ymin
xmax ymin
xmin ymax
xmax ymax
```
the order of the x/y coordinates is not as what I intended to form a ring geometry. i.e. the edges of the polygon does not cross.

Instead, I've devised the code below, whose order is as I intended. 

```
>>> for i in range(4):
...     x = list_xy_max[i // 2]
...     y = list_xy_max[int((i % 3)>0) + 2]
...     print(x, y)
... 
xmin ymin
xmin ymax
xmax ymax
xmax ymin
```

The code seems to work, but to me it's less intuitive about that it's doing. So I'm not sure which one I should choose. less boilerplate, or more readability?

src/compass/s1_cslc_qa.py
seongsujeong: good catch. I've revised the implementation based on `ogr`, `osr`, and `gdal`.

src/compass/s1_cslc_qa.py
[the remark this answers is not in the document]
LiangJYu: My suggestion was wrong. Sorry to have mistakenly sent you down this rabbit hole. The existing code is good and readable.

PR #168, opened 2023-06-02
fix `try,except` and `except, pass` codacy issues

LiangJYu:
This PR fixes #165 "codacy issues with `try, except` and `except, pass`". Codacy page [here](https://app.codacy.com/gh/opera-adt/COMPASS/pullRequest?prid=11893229).

Changes in PR:
1. added check if file exists
2. used specific exception from `jpg` crash, instead of capturing all of them.
3. removed `pass` and replaced with `gdal.OpenEx` code since that is the logical thing to follow should `gdal.Open` fail on account of `jpg` confusion.

PR #207, opened 2023-08-29
lut.py: use GTiff instead of ENVI to write scratch files

### LINKED ISSUE #206: GDAL mistakes one of the ENVI correction rasters for a JPEG and fails
@taliboliver found a very sneaky bug:
Reproduce with:
```bash
s1_cslc.py /home/staniewi/repos/COMPASS/talib-error/geo_runconfig_20230518_t042_088919_iw3.yaml
```

```python-traceback
Warning 1: libjpeg: Corrupt JPEG data: 405 extraneous bytes before marker 0x6e (this warning can be turned as an error by setting GDAL_ERROR_ON_LIBJPEG_WARNING to TRUE)
  File "/u/aurora-r0/staniewi/miniconda3/envs/mapping-311/bin/s1_cslc.py", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/s1_cslc.py", line 53, in main
    run(parser.run_config_path, parser.args.grid_type)
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/s1_cslc.py", line 45, in run
    s1_geocode_slc.run(cfg)
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/s1_geocode_slc.py", line 92, in run
    cumulative_correction_luts(burst, dem_path=cfg.dem,
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/utils/lut.py", line 63, in cumulative_correction_luts
    compute_geocoding_correction_luts(burst,
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/utils/lut.py", line 212, in compute_geocoding_correction_luts
    [open_raster(raster_path) for raster_path in rdr2geo_raster_paths]
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/utils/lut.py", line 212, in <listcomp>
    [open_raster(raster_path) for raster_path in rdr2geo_raster_paths]
     ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/utils/helpers.py", line 351, in open_raster
    ds = gdal.Open(filename, gdal.GA_ReadOnly)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/aurora-r0/staniewi/miniconda3/envs/mapping-311/lib/python3.11/site-packages/osgeo/gdal.py", line 5297, in Open
    return _gdal.Open(*args)
           ^^^^^^^^^^^^^^^^^
RuntimeError: libjpeg: Unsupported marker type 0x6e
```

Apparently the `x.rdr` raster inside the `corrections` folder is getting misread by GDAL. I've seen a similar problem with other binary drivers here https://github.com/OSGeo/gdal/issues/4713 , but I think GDAL uses a heuristic to guess the driver, and if the contents of the file match the heuristic, it can guess wrong.

scottstanie:
Fixes #206

Not sure if there are any other reasons not to use GTiff, or if this is expected to fail in other ways, but it seems to pass the unit tests and it made the reproducer run successfully.

### DIFF
--- a/src/compass/utils/lut.py
+++ b/src/compass/utils/lut.py
@@ -403,14 +403,14 @@ def compute_rdr2geo_rasters(burst, dem_raster, output_path,
                                          threshold=1.0e-8)
 
     # Get the rdr2geo raster needed for SET computation
-    topo_output = {f'{output_path}/x.rdr': gdal.GDT_Float64,
-                   f'{output_path}/y.rdr': gdal.GDT_Float64,
-                   f'{output_path}/height.rdr': gdal.GDT_Float64,
-                   f'{output_path}/incidence_angle.rdr': gdal.GDT_Float32,
-                   f'{output_path}/heading_angle.rdr': gdal.GDT_Float32}
+    topo_output = {f'{output_path}/x.rdr.tif': gdal.GDT_Float64,
+                   f'{output_path}/y.rdr.tif': gdal.GDT_Float64,
+                   f'{output_path}/height.rdr.tif': gdal.GDT_Float64,
+                   f'{output_path}/incidence_angle.rdr.tif': gdal.GDT_Float32,
+                   f'{output_path}/heading_angle.rdr.tif': gdal.GDT_Float32}
     raster_list = [
         isce3.io.Raster(fname, rdr_grid.width,
-                        rdr_grid.length, 1, dtype, 'ENVI')
+                        rdr_grid.length, 1, dtype, 'GTiff')
         for fname, dtype in topo_output.items()]
     x_raster, y_raster, height_raster, incidence_raster, heading_raster = raster_list
 


seongsujeong:
We were aware of this GDAL issue that I have briefly mentioned in the ISCE3 dev meeting while ago. That is because the first 2-3 bits of the ENV raster is the same as the ["magic byte"](https://gist.github.com/leommoore/f9e57ba2aa4bf197ebc5) of the JPEG, and GDAL attempts to load the ENVI file as JPEG.

We've fixed that issue in #164 and #168 by forcing the GDAL drive to ENVI. I think it would be great to know if the issue in #206 is from the most up-to-date version of COMPASS.

scottstanie:
I can close this if we want to keep that other fix but pick the right exception to catch. We can just add a `gdal.UseExceptions()` to helpers, and then it will raise a `RuntimeError` instead of returning None

PR #210, opened 2023-09-19
Handle data issue about missing polynomials and polygons for a burst

seongsujeong:
This PR, in association with the PR in s1-reader [here](https://github.com/isce-framework/s1-reader/pull/134), takes care of some data issues found on Sentinel-1 SAFE frame processed by IPF 002.36

- FM rate / Doppler centroid polynomials: When the polynomial list is "out of range" for the burst's sensing start / stop, it extrapolates the polynomials by extending the numbers at the border.
  - `s1-reader` has an option to turn on / off the azimuth FM rate correction in that case. In case of CSLC, we decide to do extrapolation.

- Missing polygon: In some cases of historical data, the very last burst in a subswath does not have a polygon to approximate its coverage. The new fix on `s1-reader` fills `Sentinel1Burst.border` member as an empty shapely polygon. When that happens, COMPASS skips loading that burst


vbrancat:
@seongsujeong I believe that PR #212 takes care of the issue in this PR. Do you mind closing this PR if not needed?

seongsujeong:
Closing this PR as #212 will take care of the issue.

PR #212, opened 2023-10-04
Missing polygon, fake jpeg, and turning off azimuth FM rate

seongsujeong:
This PR is an attempt to fix the issues triaged by the test by SDS.
- When the bursts's polygon is missing, the SET will be turned off. Also the center point in the metadata will be populated as [`NaN`, `NaN`]
- Unpredictable behavior was discovered from GDAL regarding error handling. Depending on GDAL's error handling mode, The fake JPEG issue will raise an error on either from `gdal.Open()`, or when attempting to access GDAL methods from `None`. The fix captures both types of error, and takes care of that.

EDIT:
- the missing polygon issue will stay as it is, so that the job gets failed
- Azimuth FM rate fill be turned off for OPERA production

### DIFF
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -545,13 +545,6 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
     for meta_item in vrt_items:
         add_dataset_and_attrs(vrt_group, meta_item)
 
-    # Fill in the center coordinated with NaN when
-    # the burst does not have center point information
-    if burst.center.is_empty:
-        xy_center = [float('nan')] * 2
-    else:
-        xy_center = [xy[0] for xy in burst.center.coords.xy]
-        
     # burst items
     burst_meta_items = [
         Meta('ipf_version', str(burst.ipf_version),
@@ -594,7 +587,7 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
         Meta('polarization', burst.polarization, 'Polarization of the burst'),
         Meta('platform_id', burst.platform_id,
              'Sensor platform identification string (e.g., S1A or S1B)'),
-        Meta('center', xy_center,
+        Meta('center', [xy[0] for xy in burst.center.coords.xy],
              'Longitude, latitude center of burst', {'units':'degrees'}),
         # window parameters
         Meta('range_window_type', burst.range_window_type,
@@ -614,7 +607,6 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
         add_dataset_and_attrs(burst_meta_group, meta_item)
 
     # Add parameters group in processing information
-    flag_set_applied = not burst.border[0].is_empty
     if save_processing_parameters:
         dry_tropo_corr_enabled = \
             True if (cfg.weather_model_file is not None) and \
@@ -646,7 +638,7 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
             Meta('geometry_doppler_applied',
                  bool(cfg.lut_params.enabled),
                  "If True, geometry steering doppler timing correction has been applied"),
-            Meta('los_solid_earth_tides_applied', bool(flag_set_applied),
+            Meta('los_solid_earth_tides_applied', bool(cfg.lut_params.enabled),
                  "If True, solid Earth tides correction has been applied in slant range direction"),
             Meta('azimuth_solid_earth_tides_applied', False,
                  "If True, solid Earth tides correction has been applied in azimuth direction"),
--- a/src/compass/utils/h5_helpers.py
+++ b/src/compass/utils/h5_helpers.py
@@ -545,6 +545,13 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
     for meta_item in vrt_items:
         add_dataset_and_attrs(vrt_group, meta_item)
 
+    # Fill in the center coordinated with NaN when
+    # the burst does not have center point information
+    if burst.center.is_empty:
+        xy_center = [float('nan')] * 2
+    else:
+        xy_center = [xy[0] for xy in burst.center.coords.xy]
+        
     # burst items
     burst_meta_items = [
         Meta('ipf_version', str(burst.ipf_version),
@@ -587,7 +594,7 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
         Meta('polarization', burst.polarization, 'Polarization of the burst'),
         Meta('platform_id', burst.platform_id,
              'Sensor platform identification string (e.g., S1A or S1B)'),
-        Meta('center', [xy[0] for xy in burst.center.coords.xy],
+        Meta('center', xy_center,
              'Longitude, latitude center of burst', {'units':'degrees'}),
         # window parameters
         Meta('range_window_type', burst.range_window_type,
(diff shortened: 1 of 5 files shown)


src/compass/utils/h5_helpers.py
```diff
@@ -607,6 +614,7 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
         add_dataset_and_attrs(burst_meta_group, meta_item)
 
     # Add parameters group in processing information
+    flag_set_applied = not burst.border[0].is_empty
```
seongsujeong, line 617: This flag gets turned on / off depending on what we have in `burst.border`. This is not a straightforward way of checking if SET was applied, but is the way I could think of to minimize the code modification. Any idea is appreciated.

src/compass/utils/h5_helpers.py
```diff
@@ -545,6 +545,13 @@ def metadata_to_h5group(parent_group, burst, cfg, save_noise_and_cal=True,
     for meta_item in vrt_items:
         add_dataset_and_attrs(vrt_group, meta_item)
 
+    # Fill in the center coordinated with NaN when
+    # the burst does not have center point information
+    if burst.center.is_empty:
+        xy_center = [float('nan')] * 2
+    else:
+        xy_center = [xy[0] for xy in burst.center.coords.xy]
+
```
vbrancat, line 554: Just a thought. Do you think there is a way to inform the users that the center is (NaN, NaN) because of missing metadata in the original input granule? I do not want the users to believe that this is the result of a glitch of our processor.

src/compass/utils/h5_helpers.py
vbrancat: I think this is fine. Can you add a comment explaining why we are doing this (please, mention the issue with S1 historical data)

src/compass/version.py
```diff
@@ -8,6 +8,7 @@
 # release history
 Tag = collections.namedtuple('Tag', 'version date')
 release_history = (
+    Tag('0.5.2', '2023-10-04'),
```
vbrancat, line 11: Not sure if you have seen the email from S. Lewis. He asked use to tag the COMPASS repository for our previous point release. Not sure how to go about this and if it is possible to tag historical states of the repository. May be for safety we should use `0.5.3`?

vbrancat (commented):
LGTM, approval pending resolution of Codacy issues. Do you have test data to test this PR? Can you point me to them? Thanks

vbrancat:
@LiangJYu can you please review this PR? It is time-sensitive. Thanks.

src/compass/version.py
seongsujeong: @vbrancat I think it's doable to tag the current main branch to `0.5.2` with the small PR for that. Do we need to deliver the Docker image to SDS after tagging it?

src/compass/utils/h5_helpers.py
seongsujeong: Considering this as resolved because we are not going to take care of missing polygon issue for SET correction.

src/compass/utils/h5_helpers.py
seongsujeong: This change was reverted. Considering this as resolved because we are not going to take care of missing polygon issue for SET correction.

src/compass/version.py
seongsujeong: Submitted another PR #213 to bump the version. #213 need to be merged before this PR to make the change to take place in order.

src/compass/utils/helpers.py
```diff
@@ -348,10 +348,17 @@ def open_raster(filename, band=1):
...
+        # In pythonic exception halding, `gdal.Open()` will raise `RuntimeError`
+        # In traditional GDAL, the function does not raise exception buy will return `None`,
+        # and the attempts to call GDAL methods will raise `AttributeError`
+
         ds = gdal.Open(filename, gdal.GA_ReadOnly)
         arr = ds.GetRasterBand(band).ReadAsArray()
         return arr
-    except AttributeError:
+
+    except (AttributeError, RuntimeError):
```
hfattahi: Since GDAL has surprised us once, shall we get to except regardless of the error type?
```suggestion
    except :
```

hfattahi (approved):
LGTM, Thanks.

src/compass/utils/lut.py
```diff
@@ -80,8 +80,9 @@ def cumulative_correction_luts(burst, dem_path, tec_path,
         rg_lut_data += dry_los_tropo
 
     # Invert signs to correct for convention
-    # TO DO: add azimuth SET to LUT
-    az_lut_data = -(bistatic_delay.data + az_fm_mismatch.data)
+    #az_lut_data = -(bistatic_delay.data + az_fm_mismatch.data)
+    az_lut_data = -(bistatic_delay.data)
```
vbrancat, line 84:
```suggestion
    az_lut_data = -bistatic_delay.data
```
Parenthesis are redundant

vbrancat (approved):
LGTM

src/compass/utils/helpers.py
```diff
@@ -348,10 +348,17 @@ def open_raster(filename, band=1):
         raise FileNotFoundError(err_str)
 
     try:
+        # The pythonic execption handling for GDAL can be turned on / off
+        # The flag of which can be identified by `gdal.GetUseExceptions()`
+        # In pythonic exception halding, `gdal.Open()` will raise `RuntimeError`
```
LiangJYu, line 353:
```suggestion
        # In pythonic exception handling, `gdal.Open()` will raise `RuntimeError`
```

src/compass/utils/helpers.py
```diff
@@ -348,10 +348,17 @@ def open_raster(filename, band=1):
...
+        # In traditional GDAL, the function does not raise exception buy will return `None`,
+        # and the attempts to call GDAL methods will raise `AttributeError`
+
         ds = gdal.Open(filename, gdal.GA_ReadOnly)
         arr = ds.GetRasterBand(band).ReadAsArray()
         return arr
-    except AttributeError:
+
+    except Exception:
```
LiangJYu, line 361:
```suggestion
    except Exception (RuntimeError, AttributeError):
```
Did the above not work? I was able to get it work with the following toy problem:
```
In [30]: def raise_one_or_other():
    ...:     rando = np.random.rand()
    ...:     if rando > 0.5:
    ...:         raise RuntimeError(f"RuntimeError: {rando}> 0.5")
    ...:     else:
    ...:         raise AttributeError(f"AttributeError: {rando} <= 0.5")

In [31]: try:
    ...:     raise_one_or_other()
    ...: except (RuntimeError, AttributeError) as err:
    ...:     print(type(err), err.args, isinstance(err, AttributeError))
    ...: 
<class 'AttributeError'> ('AttributeError: 0.1704103346669137 <= 0.5',) True

In [32]: try:
    ...:     raise_one_or_other()
    ...: except (RuntimeError, AttributeError) as err:
    ...:     print(type(err), err.args, isinstance(err, AttributeError))
    ...: 
<class 'RuntimeError'> ('RuntimeError: 0.8876243545798816> 0.5',) False

```

Geoff recommended that [`gdal.UseExceptions` be used](https://github-fn.jpl.nasa.gov/isce-3/isce/issues/922). It's used on occasion in ISCE3 and it's called immediately after `from osgeo import gdal`. If this recommendation is adopted in this file, then only the `RuntimeError` needs to be caught.

Whatever the case I would prefer catching specific errors so unexpected errors can be raised.

I forgot why we don't exclusively `gdal.OpenEx` in the `except` block and avoid `gdal.Open`?

PR #213, opened 2023-10-05
bump the version for the final point 2nd delivery

seongsujeong:
This PR is to bump the version and tag COMPASS in the final point 2nd delivery.


LiangJYu (approved):
LGTM

PR #214, opened 2023-10-13
Add azimuth FM rate into azimuth timing correction LUT

seongsujeong:
This PR is to prepare for the 3rd final point release to turn the azimuth FM rate correction `ON`, and fix the GDAL error when loading ENVI file.
Please note that the version has bumped from `0.5.3` to `0.5.4`, and `0.5.3` was not delivered to SDS.
In `0.5.3`, azimuth FM rate was turned `OFF`, and gdal issue was fixed. In this PR, the azimuth FM rate is applied again, therefore, compared to the most recent delivery (i.e. `0.5.2`), this PR only addresses the GDAL issue.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -57,7 +57,6 @@ RUN mkdir RAiDER &&\
     python -m pip install ./RAiDER &&\
     rm -rf RAiDER
 
-
 # installing OPERA COMPASS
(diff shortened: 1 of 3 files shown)


vbrancat (approved):
LGTM

LiangJYu (approved):
LGTM

PR #224, opened 2024-07-18
Fixing typos for `s1_cslc_radar` and `s1_resample`

mfangaritav:
This PR is to add the `compute_ground_to_sat_east` and `compute_ground_to_sat_north` options in the yaml schema and to fix few typos in the s1_resample.py and utils/runconfig.py for the cslc workflow.

### DIFF
--- a/src/compass/schemas/s1_cslc_radar.yaml
+++ b/src/compass/schemas/s1_cslc_radar.yaml
@@ -78,6 +78,8 @@ rdr2geo_options:
    compute_local_incidence_angle: bool(required=False)
    # Enable/disable azimuth angle output
    compute_azimuth_angle: bool(required=False)
+   compute_ground_to_sat_east: bool(required=False)
+   compute_ground_to_sat_north: bool(required=False)
 
 resample_options:
    # Lines per block to process in batch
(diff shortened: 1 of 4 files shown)


src/compass/schemas/s1_cslc_radar.yaml
```diff
@@ -78,6 +78,8 @@ rdr2geo_options:
    compute_local_incidence_angle: bool(required=False)
    # Enable/disable azimuth angle output
    compute_azimuth_angle: bool(required=False)
+   compute_ground_to_sat_east: bool(required=False)
+   compute_ground_to_sat_north: bool(required=False)
```
scottstanie, line 82: Could you add the corresponding entries in the `defaults` YAML as well?

mfangaritav: No problem, I added the options in the defaults

scottstanie (approved):

scottstanie:
Thanks for the patience @mfangaritav

jhkennedy:
@scottstanie Thank you! Just noticed this went through. Any idea when we can expect the next release?

scottstanie:
> @scottstanie Thank you! Just noticed this went through. Any idea when we can expect the next release?

I think it should happen some time in the next month- We're currently checking for any other Sentinel-1C pending changes we'll need to make to have OPERA be compatible for the new data.

PR #230, opened 2025-01-03
Fix intermediate raster format

seongsujeong:
This PR is for fixing the SAS failure reported by the SDS team in November 2024. The reason for the failure was because in rare case GDAL did not recognize the intermediate raster files written in ENVI format. The fix was to change their format from ENVI to Geotiff.

The change was tested by processing the two cases of triaged dataset from PGE, which were successful. Also the old CSLC products were re-processed with the fixed COMPASS, and they are compared by using [validate_product.py](https://github.com/opera-adt/COMPASS/blob/main/src/compass/utils/validate_product.py). The comparison did not catch any differences between them.

The increase of the speed was also tested. A slight increase in the processing time was noticed.

Burst CSLC: 
- before: 1 minutes 05 seconds
- after: 1 minutes 06 seconds

Static layer:
- before: 2 minutes 03 seconds
- after: 2 minutes 15 seconds

### DIFF
--- a/src/compass/s1_cslc.py
+++ b/src/compass/s1_cslc.py
@@ -37,7 +37,7 @@ def run(run_config_path: str, grid_type: str):
 
     elif grid_type == 'geo':
         # CSLC workflow in geo-coordinates
-        # get a runconfig dict from command line argumens
+        # get a runconfig dict from command line arguments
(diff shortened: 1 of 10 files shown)


scottstanie (approved):

scottstanie:
@seongsujeong Anything else we need to check for this?

seongsujeong:
Nothing left for this topic. I think we can merger this PR.

Issue #216, opened 2024-01-11
[New Feature]: Clarification on running workflow locally


scottyhq:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

Thanks for making this code public! We're interested in creating CLSCs at different global locations back to 2014.  

As a test I tried to recreate this file from ASF locally: `OPERA_L2_CSLC-S1_T013-026571-IW1_20231007T142223Z_20231009T193106Z_S1A_VV_v1.0.h5` 

Each HDF contains its own runconfig, but the paths to ancillary data are hardcoded and it's not clear how to get those files in place : 

```
runconfig:
  groups:
    dynamic_ancillary_file_group:
      dem_description: Digital Elevation Model (DEM) for the NASA OPERA project version
        1.1 (v1.1) based on the Copernicus DEM 30-m and Copernicus 90-m referenced
        to the WGS84 ellipsoid
      dem_file: /home/compass_user/input_dir/dem.vrt
      tec_file: /home/compass_user/input_dir/JPL0OPSRAP_20232800000_01D_02H_GIM.INX
      weather_model_file: null
    input_file_group:
      burst_id: null
      orbit_file_path:
      - /home/compass_user/input_dir/S1A_OPER_AUX_RESORB_OPOD_20231007T154954_V20231007T120659_20231007T152429.EOF
      safe_file_path:
      - /home/compass_user/input_dir/S1A_IW_SLC__1SDV_20231007T142221_20231007T142248_050660_061A8B_76E7.zip
    output:
      chunk_size:
      - 128
      - 128
      compression_enabled: true
      compression_level: 4
      cslc_data_type: complex64_zero_mantissa
      shuffle: true
    pge_name_group:
      pge_name: CSLC_S1_PGE
    primary_executable:
      product_type: CSLC_S1
    processing:
      correction_luts:
        azimuth_spacing: 0.028
        enabled: true
        range_spacing: 120
        troposphere:
          delay_type: wet_dry
      geo2rdr:
        lines_per_block: 1000
        numiter: 25
        threshold: 1.0e-08
      geocoding:
        flatten: true
        x_posting: 5
        x_snap: null
        y_posting: 10
        y_snap: null
      polarization: co-pol
      rdr2geo:
        compute_ground_to_sat_east: true
        compute_ground_to_sat_north: true
        compute_height: false
        compute_latitude: false
        compute_layover_shadow_mask: true
[...]

PR #231, opened 2025-01-28
Add `libgdal-netcdf` and `libgdal-hdf5` to `environment.yml`

scottstanie:
See https://github.com/opera-adt/COMPASS/issues/216#issuecomment-2619080450 for problem due to the GDAL split.


seongsujeong (approved):
LGTM. Thanks for the quick fix!

